PR #25, opened 2020-12-09
ci: BH-739 On tag dpeloy zip to service

markkelnar:


markkelnar:
Needs to point at other develop branch.

PR #31, opened 2020-12-09
ci: BH-739 On tag dpeloy zip to service

markkelnar:
Trigger plugin/wpe-headless with tag specified with format

```
git tag plugin/wpe-headless/0.1.1-dev 
git push --tags
```

Screen Shot 2020-12-09 at 11 20 08 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,22 @@ jobs:
             # (optional) additional CI parameters to include in the pipeline (must be a valid JSON)
             # CI_PARAMETERS: '{ "deploy": true }'
 
+  trigger-tag:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - checkout
+      - run:
+          # See MarkKelnar about the personal api token created and added as an environment variable for the project
+          name: Set token
+          command: echo "export CIRCLE_USER_TOKEN=${CIRCLE_PROJECT_TOKEN}" >> $BASH_ENV
+      - run:
+          name: Trigger workflows with tag
+          command: chmod +x .circleci/monorepo.sh && .circleci/monorepo.sh
+          environment:
+            MONOREPO_DEBUG: "true"
+            TRIGGER_PACKAGE: wpe-headless-plugin
+
   checkout:
     executor: wp-product-orb/default
     steps:
@@ -161,6 +177,14 @@ workflows:
     when: << pipeline.parameters.trigger >>
     jobs:
       - trigger-workflows
+
+  tag:
+    when: << pipeline.git.tag >>
+    jobs:
+      - trigger-tag:
+          filters:
+            tags:
+              only: /^plugin\/wpe-headless\/\S+/
 
   # Workflows defined for each package.
   plugin-wpe-headless:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,8 +179,12 @@ workflows:
     jobs:
       - trigger-workflows
 
+  # Trigger tag build when tag and not already building a specific package workflow
   tag:
-    when: << pipeline.git.tag >>
+    when: 
+      and:
+        - << pipeline.git.tag >>
+        - not: << pipeline.parameters.wpe-headless-plugin >>
     jobs:
       - trigger-tag:
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,11 +128,16 @@ jobs:
           paths:
             - plugins/wpe-headless/build
 
-  plugin-wpe-headless-deploy-staging:
+  plugin-wpe-headless-deploy:
     executor: wp-product-orb/authenticate
     working_directory: ~/project/plugins/wpe-headless
     environment:
       WPE_SESSION_DIR: ./.wpe
+    parameters:
+      auth_url:
+        type: string
+      upload_url:
+        type: string
     steps:
       - attach_workspace:
           at: ~/project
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,6 +37,23 @@ jobs:
             # (optional) additional CI parameters to include in the pipeline (must be a valid JSON)
             # CI_PARAMETERS: '{ "deploy": true }'
 
+  trigger-tag:
```
markkelnar, line 40: Add a job specific to invoke the monorepo script.

.circleci/config.yml
```diff
@@ -37,6 +37,23 @@ jobs:
...
+          # See MarkKelnar about the personal api token created and added as an environment variable for the project
+          name: Set token
+          command: echo "export CIRCLE_USER_TOKEN=${CIRCLE_PROJECT_TOKEN}" >> $BASH_ENV
+      - run:
+          name: Trigger workflows with tag
+          command: chmod +x .circleci/monorepo.sh && .circleci/monorepo.sh
+          environment:
+            MONOREPO_DEBUG: "true"
+            # (optional) additional CI parameters to include in the pipeline (must be a valid JSON)
+            CI_PARAMETERS: '{ "wpe-headless-plugin": true }'
```
markkelnar, line 55: We only have the one plugin/package at the moment that needs this behavior. Can come back and refine this to be more flexible but trying to get this out the door.

.circleci/config.yml
```diff
@@ -157,28 +179,75 @@ workflows:
     jobs:
       - trigger-workflows
 
+  # Trigger tag build when tag and not already building a specific package workflow
+  tag:
+    when:
```
markkelnar, line 184: This workflow specific for when is tag is present.

.circleci/config.yml
```diff
@@ -157,28 +179,75 @@ workflows:
...
+          name: "Deploy zip to api (production)"
           requires:
             - plugin-wpe-headless-bundle-zip
             - plugin-wpe-headless-bundle-json
           filters:
             branches:
-              only: main
+              ignore: /.*/
+            tags:
+              # tag ex. plugin/wpe-headless/1.0.0
```
markkelnar, line 249: The tag format I came up with to specify the plugin.

.circleci/monorepo.sh
```diff
@@ -147,11 +147,19 @@ function print_status {
...
-    { branch: $branch, parameters: . } | 
+  echo "$1" |
+  jq --raw-output --arg branch "${CIRCLE_BRANCH}" --arg trigger "${TRIGGER_PARAM_NAME}" --argjson params "${CI_PARAMETERS:-null}" '. |
+    map(select(.changes > 0)) |
+    reduce .[] as $i (($params // {}) * { ($trigger): false }; .[$i.package] = true) |
+    { branch: $branch, parameters: . } |
+    @json'
+}
+
+function create_request_tag {
```
markkelnar, line 158: Added this function specific to building the json blog that is submitted to trigger the circleci job. Needed 'tag' instead of branch.

rfmeier (approved):

.circleci/monorepo.sh
```diff
@@ -201,11 +209,15 @@ function get_builds {
 
 function debug {
   echo -e "\n\nDEBUG INFORMATION"
-  echo -e "\n\n=== Branches ==="
-  cat "${TMP_DIR}/branches.txt"
+  if [[ -r "${TMP_DIR}/branches.txt" ]]; then
```
markkelnar, line 212: Saw errors cat'ing these files when debug was enabled and specified tag, cause these files are created.

.circleci/config.yml
```diff
@@ -157,28 +179,75 @@ workflows:
...
+              only: /.*/
+      - plugin-wpe-headless-deploy:
+          name: "Deploy zip to api (staging)"
+          requires:
+            - plugin-wpe-headless-bundle-zip
+            - plugin-wpe-headless-bundle-json
+          filters:
+            branches:
+              only:
+                - main
```
markkelnar, line 234: @rfmeier asked if this should be canary merge.  It could be both. :looking: at that now.

.circleci/config.yml
```diff
@@ -157,28 +179,75 @@ workflows:
...
+          filters:
+            tags:
+              only: /^plugin\/wpe-headless\/\S+/
+
   # Workflows defined for each package.
   plugin-wpe-headless:
     when: << pipeline.parameters.wpe-headless-plugin >>
     jobs:
-      - checkout
+      - checkout:
```
markkelnar, line 198: These are all added, cause by default circleci doesn't trigger on tag event, so when we want to, we have to add this to all jobs in the workflow that we want responding to the tag.

claygriffiths: Interesting! Might be worth inline-comomenting this particular tidbit.

.circleci/config.yml
claygriffiths: Yeah, I think we'll want both `canary` and `main` here.

.circleci/monorepo.sh
```diff
@@ -226,25 +238,29 @@ function get_parent {
 
 function main {
   init
-  get_builds
-  git_parent_commit=$( get_parent )
+  if [[ "x${CIRCLE_TAG}" == "x" ]]; then
```
claygriffiths, line 241: Haha, this is a clever way of checking for an empty string.

claygriffiths (approved):
Nice work!

Approval pending adding support for deploying off of both `main` and `canary`.

PR #39, opened 2020-12-15
ci: simplify CircleCI config to always run wpe-headless checks

claygriffiths:
While well-intentioned, the CircleCI monorepo example that we were using caused more trouble than it helped.

This PR removes the monorepo script and instead, always checks out and runs the test for the WordPress plugin. Deployments still require a tag.


.circleci/config.yml
```diff
@@ -172,70 +126,43 @@ jobs:
...
   # Workflows defined for each package.
-  # By default circleci doesn't trigger on a tag event. Add the filter for tags so that all jobs in the workflow will run also when a tag is present.
+  # tag example for deploying an update for wpe-headless: plugin/wpe-headless/1.0.0
   plugin-wpe-headless:
-    when: << pipeline.parameters.wpe-headless-plugin >>
     jobs:
-      - checkout:
-          filters:
-            tags:
-              only: /.*/
```
markkelnar, line 202: You need these filters for any tag that way the jobs will run without a tag being specified. Ie, when a PR on a branch is created.  As well as when the actual tag is present.

.circleci/config.yml
```diff
@@ -172,70 +126,43 @@ jobs:
...
+              only: /^plugin\/wpe-headless\/\S+/
       - plugin-wpe-headless-deploy:
           name: "Deploy zip to api (staging)"
           requires:
             - plugin-wpe-headless-bundle-zip
             - plugin-wpe-headless-bundle-json
           filters:
             branches:
-              only:
```
markkelnar, line 234: This was here to only push/[ublish/deploy the zip file on merge, not every competing pr and branch constantly pushing zip files of different flavors to the staging endpoint.  That way staging represented at least what was in develop in 'canary'.

.circleci/config.yml
```diff
@@ -172,70 +126,43 @@ jobs:
...
-            tags:
-              only: /.*/
       - plugin-wpe-headless-bundle-zip:
           requires:
             - plugin-wpe-headless-checkout
           filters:
+            branches:
+              ignore: /.*/
             tags:
-              only: /.*/
```
markkelnar, line 220: Why are you changing this behavior?

claygriffiths: Since this isn't run using the trigger workflow anymore, my thought is that it should be triggered at the same time as `plugin-wpe-headless-deploy`

markkelnar: Ok, see my comment below about deploying a zip file to staging api when main or canary.

markkelnar:
I see this is targeted at main branch. Will that be rolled back to canary?

claygriffiths:
@markkelnar Great catch. 🤦‍♂️  I'll update the base now.

.circleci/config.yml
```diff
@@ -284,7 +210,9 @@ workflows:
             - plugin-wpe-headless-bundle-json
           filters:
             branches:
-              ignore: /.*/
```
claygriffiths, line 287: @markkelnar I made this change to get it in sync with the staging job above. Does this look good to you?

markkelnar: This job will only run when the matching tag is present in the pipeline. I don't know what the branch will be and previously, I've always used the ignore * filter for branches, per circleci documentation.

markkelnar (approved):

PR #43, opened 2020-12-16
fix: Improve GraphQL authentication return type consistency.

mindctrl:
fix: Ensure existing `$authentication_errors` value is returned to GraphQL's `graphql_authentication_errors` filter when a `HTTP_AUTHORIZATION` header does not exist.

Also ensures consistent boolean return type to the filter from `wpe_headless_rest_validate_access_token()` function.

### DIFF
--- a/plugins/wpe-headless/includes/rest/callbacks.php
+++ b/plugins/wpe-headless/includes/rest/callbacks.php
@@ -18,16 +18,18 @@ add_filter( 'graphql_authentication_errors', 'wpe_headless_rest_validate_access_
  *
  * @link https://www.wpgraphql.com/filters/graphql_authentication_errors/
  *
+ * @param bool $authentication_errors Whether there are authentication errors with the request.
+ *
  * @return bool True if the Authorization header exists and is invalid, false otherwise.
  */
-function wpe_headless_rest_validate_access_token() {
+function wpe_headless_rest_validate_access_token( $authentication_errors ) {
 	if ( ! isset( $_SERVER['HTTP_AUTHORIZATION'] ) ) {
-		return false;
+		return $authentication_errors;
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/wpe-headless/includes/rest/callbacks.php
```diff
@@ -18,24 +18,26 @@
  *
  * @link https://www.wpgraphql.com/filters/graphql_authentication_errors/
  *
+ * @param bool $authentication_errors Whether there are authentication errors with the request.
+ *
  * @return bool True if the Authorization header exists and is invalid, false otherwise.
  */
-function wpe_headless_rest_validate_access_token() {
+function wpe_headless_rest_validate_access_token( $authentication_errors ) {
```
mindctrl, line 25: I wonder if we can improve the name of this function. With a return value of `true` meaning something like "there are authentication errors", it doesn't quite fit the name. I would expect a function named `wpe_headless_rest_validate_access_token` to return `true` if the token is valid.

mindctrl: Maybe something like `wpe_headless_graphql_authentication_errors` or similar?

claygriffiths: Dang, good Q. It is somewhat awkward.

Perhaps something like `wpe_headless_graphql_has_auth_errors`? I'm a fan of predicates having `is` or `has` in their name.

claygriffiths (approved):
Thanks for fixing this up! ✅ as the code looks good

rfmeier (approved):

mindctrl:
@rfmeier @claygriffiths thanks for the feedback. You both make good points. Ryan, the distinction between a utility function and a callback is a good one. For now I'm going to merge this as is. We can deal with a rename later when polishing things up like Trevan mentioned earlier.

PR #51, opened 2020-12-22
BH-756 Adds one-click install for WP GraphQL plugin.

mindctrl:
This PR adds:
- Redirect to settings page when plugin is activated.
- Checks if WP GraphQL plugin is installed.
- Admin notice if GraphQL plugin is not installed.
- Functionality to install and activate GraphQL plugin from admin notice.

Most of this code is meant to be deleted when we have the setup wizard, and it's written as such. The JS strings aren't translatable. The JS isn't in a separate file. It's not transpiled to work in every browser. It's not trying to catch every edge case error that could happen. Happy to improve upon those things if we think it's necessary.

### DIFF
--- a/plugins/wpe-headless/includes/settings/callbacks.php
+++ b/plugins/wpe-headless/includes/settings/callbacks.php
@@ -196,3 +196,140 @@ function wpe_headless_display_frontend_uri_field() {
 	</p>
 	<?php
 }
+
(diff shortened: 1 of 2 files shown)


nickcernis (approved):
Looking good. Confirmed that:
- Redirect occurs on activation.
- Plugin installation/activation completes.
- No admin notice if WPGraphQL is active.

PR #52, opened 2020-12-22
Enable secret key regeneration

rfmeier:
This pull request allows a user to regenerate the secret key. Better copy is welcome.

Screen Shot 2020-12-22 at 10 51 06 AM

A confirmation will be presented to the user allowing the user to prevent secret key regeneration if needed.

Screen Shot 2020-12-22 at 10 51 36 AM


nickcernis (approved):
Working great. Nice implementation. I think the copy is fine for now — it's probably enough to ask for confirmation without explaining the impact in a dialog.

PR #54, opened 2021-01-04
@wpengine/headless package housekeeping

claygriffiths:
* Add some scripts in `package.json` for DX
* Add `types` to `package.json`
* Update root for TypeScript so we don't wind up with `dist/src`.
* Remove `react` and `react-dom` from devDeps to help reduce conflicts when working on the package
* Remove explicit ports on `dev` and `start` in our `preview` example

### DIFF
--- a/packages/headless/tsconfig.json
+++ b/packages/headless/tsconfig.json
@@ -1,29 +1,29 @@
 {
-    "compilerOptions": {
-        "module": "CommonJS",
-        "moduleResolution": "Node",
-        "target": "ES2017",
-        "removeComments": true,
-        "allowSyntheticDefaultImports": true,
-        "skipLibCheck": true,
-        "esModuleInterop": true,
-        "strict": true,
-        "noUnusedLocals": true,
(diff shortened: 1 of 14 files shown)


packages/headless/tsconfig.json
```diff
@@ -10,19 +10,20 @@
...
         ],
         "jsx": "react"
     },
-    "include": [
-        "**/*.ts",
-        "**/*.tsx"
-    ],
+	"include": [
+		"src"
+	],
```
mindctrl, line 24:
```suggestion
    "include": [
        "src"
    ],
```

packages/headless/tsconfig.json
```diff
@@ -10,19 +10,20 @@
...
     "exclude": [
-        "node_modules"
+        "node_modules",
+		"dist"
```
mindctrl, line 27:
```suggestion
        "dist"
```

mindctrl (approved):
Minor ident issues, otherwise LGTM.

PR #55, opened 2021-01-04
feat: add support for Next.js SSG + Previewing

claygriffiths:
# Summary

This PR adds support for Next.js Static Site Generation (SSG) + WordPress Previewing. All of the changes are compatible with our existing previewing examples/articles that utilize server-side rendering (SSR).

Because SSG is superior in terms of performance and the recommended Data Fetching paradigm for Next.js, the preview example in `examples/preview` has been updated to utilize SSG.

# Technical

* Allow `WORDPRESS_URL` environment variable to only be set for server-side if desired
* Register `conditionalTags` and `templates` fields to the `ContentNode` type in WP GraphQL. Previously, we only registered it on `UniformResourceIdentifiable`
* Add new `wpe_headless_post_preview_link` callback to the WordPress plugin which handles redirecting to `api/auth/wpe-headless` by default rather than the permalink on the frontend. This is required as visiting a statically rendered page on Next.js doesn't provide an opportunity to set Next.js `previewMode` cookies. 
* Update `initializeNextServerSideProps` to use Next.js preview mode much like `initializeNextStaticProps`

### DIFF
--- a/examples/preview/pages/404.tsx
+++ b/examples/preview/pages/404.tsx
@@ -1,0 +1,6 @@
+/* eslint-disable react/jsx-props-no-spreading */
+import React from 'react';
+
+export default function Page404() {
+  return <h1>404</h1>;
+}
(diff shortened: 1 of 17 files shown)


wjohnsto (approved):
Looks good!

PR #58, opened 2021-01-04
WordPress Plugin Feature Toggling

rfmeier:
This pull request adds three settings to allow users to enable or disable plugin features.

- Disable WordPress theme functionality
  - Removes theme related links.
- Enable Post and Category URL rewrites
  - Enable or disable content, and taxonomy url rewrites
- Enable public route redirects
  - Enable or disable front-end redirects

When the plugin is activated for the first time, the three settings will be enabled (only if the settings do not exist in the database).

The copy/naming used may be poor. Let me know if any changes are needed.

Screen Shot 2021-01-04 at 1 53 02 PM

### DIFF
--- a/plugins/wpe-headless/includes/admin-menus/callbacks.php
+++ b/plugins/wpe-headless/includes/admin-menus/callbacks.php
@@ -14,6 +14,7 @@ add_action( 'admin_menu', 'wpe_headless_remove_admin_menu_pages', 1000 );
  * Remove wp-admin menu items not needed in a headless environment.
  *
  * @global $submenu
+ *
  * @return void
  */
 function wpe_headless_remove_admin_menu_pages() {
@@ -23,6 +24,12 @@ function wpe_headless_remove_admin_menu_pages() {
 	 * @var array
 	 */
 	global $submenu;
+
+	$disable_theme = wpe_headless_get_setting( 'disable_theme', false );
+
+	if ( ! $disable_theme ) {
+		return;
+	}
 
 	// Remove Appearance > Themes.
 	remove_submenu_page( 'themes.php', 'themes.php' );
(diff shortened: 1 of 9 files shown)


plugins/wpe-headless/includes/admin-menus/callbacks.php
```diff
@@ -24,6 +25,12 @@ function wpe_headless_remove_admin_menu_pages() {
 	 */
 	global $submenu;
 
+	$disable_theme = wpe_headless_get_setting( 'disable_theme', false );
```
claygriffiths, line 28: Curious, what's the rationale behind adding the `false` here? I see that it's the default value for the argument.

I see that the value is set to `1` on activation. If it's unchecked, does the value get removed from the settings array entirely?

claygriffiths (approved):
Good stuff!

plugins/wpe-headless/includes/admin-menus/callbacks.php
rfmeier: >I see that the value is set to 1 on activation. If it's unchecked, does the value get removed from the settings array entirely?

Yes, unchecked and the value will be removed from the settings, which is why the default is `false` and not `true`. Else, it would constantly return true.

rfmeier: And `false` is the default value for the `$default` parameter in `wpe_headless_get_setting()`, I essentially could just remove the second parameter.

claygriffiths: Got it, makes sense! I like how it's explicit. I suppose my only feedback here would be to have a getter function for all of these toggles like you already do for `wpe_headless_domain_replacement_enabled()`.

mindctrl (changes requested):
@rfmeier looks like "Enable public route redirects" isn't wired into `wpe_headless_deny_public_access()`, so redirects always happen regardless of the setting.

rfmeier:
>@rfmeier looks like "Enable public route redirects" isn't wired into wpe_headless_deny_public_access(), so redirects always happen regardless of the setting.

@mindctrl Thanks for catching that. Not sure how it was working before...

mindctrl (approved):

mindctrl (approved):

PR #68, opened 2021-01-12
Static Block Styling

rfmeier:
This pull request will add styling for static WordPress blocks.

1. Added graphql query to return stylesheet information.
2. Add stylesheet url for WordPress Block Library if exists.


packages/headless/src/components/WPHead.tsx
```diff
@@ -17,9 +24,25 @@ export default function WPHead(): JSX.Element {
     title = `${siteTitle} – ${siteTagline}`;
   }
 
+  if (post?.enqueuedStylesheets?.nodes) {
+    stylesheet = post.enqueuedStylesheets.nodes
+      .filter((node) => {
+        return node.handle === 'wp-block-library';
```
claygriffiths, line 30: Shouldn't we include all enqueued stylesheets in case other blocks enqueue additional styles?

I know that disabling the theme in some capacity is planned, but as a stop-gap, perhaps we could filter out any URLs that start with `wp-content/themes`? 🤔

packages/headless/src/components/WPHead.tsx
```diff
@@ -1,12 +1,19 @@
 import React from 'react';
 import Head from 'next/head';
 import { usePost, useGeneralSettings } from '../api';
+import { EnqueuedStylesheet } from '../types';
+import { trimTrailingSlash } from '../utils';
+
+const WP_URL = trimTrailingSlash(
```
claygriffiths, line 7: This is ripe for a helper/util 😁

claygriffiths (changes requested):
I believe we'll want to enqueue all stylesheets for a given page in case there are non-core blocks that need styles or potentially shortcodes.

Other than that, nice work digging into this!

packages/headless/src/components/WPHead.tsx
claygriffiths: `WP_URL`, that is

rfmeier: Yeah. It was used in three other places before I added that.

packages/headless/src/components/WPHead.tsx
rfmeier: Updated the filter loop include everything except theme stylesheets.

packages/headless/src/components/WPHead.tsx
rfmeier: Can we add a dedicated function for this with a separate PR?

nickcernis (approved):

mindctrl (approved):

packages/headless/src/components/WPHead.tsx
claygriffiths: Yeah, totally! Sorry, didn't mean for it to be a blocker.

claygriffiths (approved):

mindctrl:
Noting that Ryan and I found a quirk with CSS from plugins. More commits incoming.

mindctrl:
Latest commits fix plugin CSS issue. LGTM now. 👍

PR #69, opened 2021-01-14
build: start leveraging lerna

claygriffiths:
Running `npm run dev` in `packages/headless` and `examples/preview` while also managing the `npm link`'s is quite cumbersome.

Fortunately, we already have Lerna in the repository which is here to solve this exact problem!

This PR sets the project up so you can work on `@wpengine/headless` with the following scripts ran in the root of the repository:

1. `npm run bootstrap`
    This will run an `npm install` as well as bootstrap Lerna which will link all of the packages in the repository together.

2. `npm run dev`
   Running this script will kick off dev/watch scripts for `@wpengine/headless`, wait for it to finish the first build, and then start up Next.js dev mode in `examples/preview`

---

image

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -17,7 +17,8 @@ When working on the NPM packages in this repository, we recommend utilizing our 
 To get going, you can run the following:
 
 1. `npm run bootstrap`
-2. `npm run dev`
+2. Ensure that `.env.local` is properly configured in `examples/preview`
+3. `npm run dev`
 
 ### Plugins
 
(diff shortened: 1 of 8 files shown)


matt-landers (approved):

mindctrl:
If you run this on a fresh clone, the build fails because of the `WPE_HEADLESS_SECRET` not being defined in `.env.local`, which won't exist because it's `.env.local.sample`. Then if you fix that, you also have to make sure `NEXT_PUBLIC_WORDPRESS_URL` or it will fail again. Just noting because the docs don't mention the need to do the config first.

Otherwise looks good!

mindctrl (approved):

PR #71, opened 2021-01-14
feat: adding featured images to gql calls

wjohnsto:
Adds featured images to WPGraphQL calls for posts.

### DIFF
--- a/packages/headless/src/types.ts
+++ b/packages/headless/src/types.ts
@@ -100,6 +100,28 @@ export interface PageInfo {
 }
 
 /**
+ * A WordPress featured image object obtained through a WPGraphQL request
+ *
+ * @export
+ * @interface FeaturedImageNode
+ */
+export interface FeaturedImageNode {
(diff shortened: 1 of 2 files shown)


packages/headless/src/types.ts
```diff
@@ -99,6 +99,28 @@ export interface PageInfo {
...
+/**
+ * A WordPress featured image node object obtained through a WPGraphQL request
+ *
+ * @export
+ * @interface FeaturedImage
+ */
+export interface FeaturedImage {
+  node: FeaturedImageNode;
+}
+
```
claygriffiths, line 123: Thanks for adding these. Just a heads-up, I'm going to see if we can automatically generate these from the GraphQL schema.

claygriffiths (approved):

packages/headless/src/types.ts
wjohnsto: That would be excellent!

PR #72, opened 2021-01-14
feat(BH-806): add middleware support for Next.js Data Fetcher in templates

claygriffiths:
# Summary

Templates in the `theme` directory can now export a function named `getPropsMiddleware`. This method will be called during either `getStaticProps()` or `getServerSideProps()` of the page calling the template loader.

The middleware makes it possible to request additional data when the Next.js Data Fetchers are run or to modify the default data being requested.

### DIFF
--- a/packages/headless/src/utils/resolveTemplate.ts
+++ b/packages/headless/src/utils/resolveTemplate.ts
@@ -1,0 +1,39 @@
+import { UriInfo } from '../types';
+import type { Template } from '../components/TemplateLoader';
+
+export function resolveTemplate(
+  pageInfo: UriInfo | undefined,
+): null | Template {
+  /**
+   * List out files in main project using Webpack.
+   */
+  let context;
+
+  try {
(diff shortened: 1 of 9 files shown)


packages/headless/src/utils/resolveTemplate.ts
```diff
@@ -0,0 +1,39 @@
...
+): null | Template {
+  /**
+   * List out files in main project using Webpack.
+   */
+  let context;
+
+  try {
+    context = require.context('theme', false, /.*/);
+  } catch (e) {
+    console.warn(
```
nickcernis, line 15: Is it worth allowing console warnings and errors to suppress this ESLint warning?

https://eslint.org/docs/rules/no-console#options

claygriffiths: Great callout! I'll make that adjustment as we'll likely continue raising warnings/errors.

nickcernis (approved):

PR #75, opened 2021-01-15
fix: Updated logic to enable content rewrites

rfmeier:
This pull request adds an additional check to `wpe_headless_domain_replacement_enabled()` to determine if the incoming request is a GraphQL request.

### DIFF
--- a/plugins/wpe-headless/includes/replacement/functions.php
+++ b/plugins/wpe-headless/includes/replacement/functions.php
@@ -20,11 +20,11 @@ function wpe_headless_domain_replacement_enabled() {
 	$enabled = false;
 
 	if ( wpe_headless_is_rewrites_enabled() ) {
-		if ( isset( $_GET['replace-domain'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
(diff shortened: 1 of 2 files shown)


plugins/wpe-headless/includes/replacement/functions.php
```diff
@@ -20,11 +20,11 @@ function wpe_headless_domain_replacement_enabled() {
 	$enabled = false;
 
 	if ( wpe_headless_is_rewrites_enabled() ) {
-		if ( isset( $_GET['replace-domain'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
nickcernis, line 23: The removed `phpcs:ignore` comment seems to be the cause of the CI failure here.

nickcernis (approved):

mindctrl (approved):
LGTM. Is this separate from BH-814 and wp-admin links?

rfmeier:
>LGTM. Is this separate from BH-814 and wp-admin links?

This is for BH-814. However, now that I get what the ticket was about, I will most likely delete and re-create this PR.

PR #78, opened 2021-01-20
docs: Improve secret key field description

nickcernis:
- Makes link text more accessible (screenreaders now say “headless post previews” instead of “here”).
- Adds missing `__()` translation function call.

### Before
Screenshot 2021-01-20 at 11 23 52

### After
Screenshot 2021-01-20 at 11 23 05

### DIFF
--- a/plugins/wpe-headless/includes/settings/callbacks.php
+++ b/plugins/wpe-headless/includes/settings/callbacks.php
@@ -229,8 +229,8 @@ function wpe_headless_display_secret_key_field() {
 	<p class="description">
 		<?php
 		printf(
-			/* translators: %s: link to documentation */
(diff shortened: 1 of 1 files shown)


plugins/wpe-headless/includes/settings/callbacks.php
```diff
@@ -229,8 +229,8 @@ function confirm_regenerate_key( event ) {
 	<p class="description">
 		<?php
 		printf(
-			/* translators: %s: link to documentation */
-			wp_kses_post( 'This key is used to enable post previewing with Next.js. Read about post previewing <a href="%s" target="_blank" rel="noopener noreferrer">here</a>.', 'wpe-headless' ),
```
nickcernis, line 233: IntelliJ was flagging this as "Method call is provided 2 parameters, but the method signature uses 1 parameters" in PhpStorm/IntelliJ, due to `'wpe-headless'` being passed as a second param to `wp_kses_post`.

mindctrl (approved):

PR #82, opened 2021-01-20
Change Next to start on port 8080

bengreeley:
Since the WP Atlas hosting/deployment service expects the service to run on port 8080, we should update the example so it's ready to work with the deployment service.


mindctrl:
@bengreeley thanks for the PR. The ports were recently removed [here](https://github.com/wpengine/headless-framework/pull/54), but I don't remember why. @claygriffiths might remember.

claygriffiths:
Thanks, @bengreeley!

PR #85, opened 2021-01-22
docs: Adjust “disable theme” label

nickcernis:
To make it more clear that the “disable theme” plugin option at Settings → Headless currently disables theme admin pages, not full theme functionality.

As discussed in Slack with @trevanhetzel and @mindctrl.

## Before
Screenshot 2021-01-22 at 11 25 01

## After
Screenshot 2021-01-22 at 11 24 44


rfmeier (approved):
This is much more clearer. Thanks, Nick.

mindctrl (approved):

PR #86, opened 2021-01-22
fix(BH-816): make redirect in initializeNextStaticProps dynamic

claygriffiths:
# Summary

`initializeNextStaticProps` has a redirect in it to handle authorization for previewing. Before this PR, it was hard-coded as there wasn't a good way of getting the current host from the static props context.

To get around the limitation mentioned above, this PR sets some info in the Next.js preview data that can be used in the static props data fetcher.


mindctrl (approved):
LGTM

PR #87, opened 2021-01-22
feat: add graphql-codegen

claygriffiths:
# Summary

Considering we're using both TypeScript and GraphQL which provide types, it makes sense to generate TypeScript types from the source of truth: the WP GraphQL schema.

TypeScript types are generated for both the entire schema and operations (queries/mutations) found throughout the `packages/headless` codebase.

## Development Notes

* To regenerate types, run `npm run generate-graphql` in `packages/headless`.
* The generated types are checked into VCS
* Writing return types for hooks and fetchers that use GraphQL should no longer be necessary as they will be generated. You may need to write the operation, run `npm run generate-graphql`, and then set the return type accordingly. See `usePost()` and `getPosts()` as examples.
* `graphql-codegen` pulls from `wpgraphql-schema.graphql` in the root of the repository


nickcernis (approved):
Nice idea! Maybe also add these notes to `DEVELOPMENT.md`? (Not blocking.)

PR #88, opened 2021-01-22
feat: Updated for WordPress plugin release.

rfmeier:
Updated versions for WordPress plugin release.


claygriffiths (approved):
LGTM

As an aside, we need to definitively decide on if we want to keep the plugin version in sync with the `@wpengine/headless` package. If so, we'll likely want to create a script that can bump the version.

PR #90, opened 2021-01-25
Add callback for content src= replacement

markkelnar:
Search the post content for src="http://frontend.com" replace with src="http://backend.com".


mindctrl (approved):
Looks good to me

mindctrl:
Note: we need `srcset` too, but can be another PR.

PR #98, opened 2021-01-26
fix: do not export resolveTemplate in /utils to reduce chance of Webpack error/warning

claygriffiths:
This should prevent errors if the 'theme' directory is not present in a non-Next.js project


nickcernis (approved):
Nice catch!

PR #99, opened 2021-01-27
fix: Ensure links are rewritten correctly in the preview app

nickcernis:
Fixes WP-to-frontend app URL replacement in GraphQL responses by:

- Matching on `href` keys as well as `src` keys in the WPGraphQL response, so that the [new menu URLs](https://github.com/wpengine/headless-framework/pull/96/files#diff-4648c0eaa39a027be656454a4ed56421c393fbf535223253c26a81438728e7bcR318-R325) will be rewritten.
- Appending `replace-domain` to the WPGraphQL path in the preview app, so that domain replacement takes place as soon as someone ticks “Enable Post and Category URL rewrites” in Settings → Headless. (The replacement logic currently requires [the checkbox to be enabled and a query string/header to be set](https://github.com/wpengine/headless-framework/blob/04a5a25d09710a68d06c15c74fbd44982afba853/plugins/wpe-headless/includes/replacement/functions.php#L22-L23) before replacement happens.)


packages/headless/src/provider/apolloClient.ts
```diff
@@ -52,7 +52,7 @@ function createApolloClient(): ApolloClient<NormalizedCacheObject> {
     link: WP_URL
       ? authLink.concat(
           new BatchHttpLink({
-            uri: `${WP_URL}/graphql`,
+            uri: `${WP_URL}/graphql?replace-domain=1`,
```
mindctrl, line 55: Just a note that we're discussing this and possibly removing the requirement to pass this parameter to the GraphQL endpoint for URL rewrites. We have a setting for enabling/disabling URL rewrites in content, and we think that's sufficient unless customers tell us otherwise.

nickcernis: Noting for the record that we dropped the `replace-domain` query string requirement in https://github.com/wpengine/headless-framework/pull/112.

I'll amend this PR to cover the href rewriting only.

nickcernis:
Updated to replace `href` attributes with the frontend URL.

Drops the `?replace-domain=1` query string, which we no longer use.

mindctrl (approved):

PR #103, opened 2021-01-27
refactor: generate wpgraphql.d.ts instead of wpgraphql.ts

claygriffiths:
Instead of generating an actual TypeScript file (`.ts`) for the WPGraphQL types, generate a type declaration file (`.d.ts`). 

The primary reason for this move was to reduce the size of the initial bundle. (Note the 10KB reduction below)

_Before_
Screen Shot 2021-01-27 at 2 38 58 PM

_After_
Screen Shot 2021-01-27 at 2 53 10 PM

---

## Other benefits

* Enables using `WPGraphQL` ambient types in the framework without importing the types everywhere they're needed. 
* Enums are now string unions, which are easier to use as a developer and do not contribute to bundle size. An example here is simply needing to use `'URI'` as a parameter instead of `WPGraphQL.ContentNodeIdTypeEnum.Uri`

### DIFF
--- a/packages/headless/.eslintignore
+++ b/packages/headless/.eslintignore
@@ -3,4 +3,4 @@ node_modules
 /*.js
 /*.ts
 /*.d.ts
-src/types/wpgraphql.ts
+src/types/wpgraphql.d.ts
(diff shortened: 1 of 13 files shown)


nickcernis (approved):
> Enums are now string unions, which are easier to use as a developer and do not contribute to bundle size. An example here is simply needing to use 'URI' as a parameter instead of WPGraphQL.ContentNodeIdTypeEnum.Uri

Nice benefit! Much better to be able to pass 'PRIMARY' instead of `WPGraphQL.MenuLocationEnum.Primary` for the menu location filter.

PR #105, opened 2021-01-28
fix: properly hydrate Apollo cache if using a static frontpage

claygriffiths:
Run `getContentNode()` if a static front page is enabled. Note, it needs to be conditionally run as the query will result in a 500 error if a static front page is not in use.

Additionally, this DRYs things up by extracting out the querying in the Data Fetchers into a new method named `nextFetchFromWP`

### DIFF
--- a/examples/preview/theme/page.tsx
+++ b/examples/preview/theme/page.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import { usePost, WPHead } from '@wpengine/headless';
+
+export default function Page() {
+  const post = usePost();
+
+  return (
+    <>
+      <WPHead />
+
+      <div>
+        {post && (
+          <div>
+            <div>
+              <h5>{post.title}</h5>
+              <div dangerouslySetInnerHTML={{ __html: post.content ?? '' }} />
+            </div>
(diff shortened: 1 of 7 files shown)


examples/preview/theme/page.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <>
+      <WPHead />
+
+      <div>
+        {post && (
+          <div>
+            <div>
+              <h5>{post.title}</h5>
```
nickcernis, line 15: Should we use h1 for the page title?

```suggestion
              <h1>{post.title}</h1>
```

(I know it's h5 in the post template too, but we can correct it separately there.)

nickcernis (approved):
Thanks for fixing this!

- [x] Confirmed `page.tsx` now shows the static front page if set at Settings → Reading.
- [x] Confirmed `front-page.tsx` can be used as a front-page specific page template.

examples/preview/theme/page.tsx
claygriffiths: Nice callout. I'll patch that up.

PR #108, opened 2021-01-29
Cookies will only be persistent for a single request

wjohnsto:
Prior to this PR there is a chance that cookies would persist across multiple requests, this corrects that.


claygriffiths (approved):

claygriffiths (changes requested):
Actually, looks like some adjustments are needed in the example.

Seeing the following:

image

claygriffiths (approved):

PR #111, opened 2021-02-01
Ci: package headless build

markkelnar:
Add initial jobs for build and test of the npm headless package.
Notice an existing test that fails and should be fixed, but considering that outside the scope of this work.


rfmeier (approved):

.circleci/config.yml
```diff
@@ -175,6 +175,37 @@ jobs:
...
+    docker:
+      - image: cimg/node:lts
+    working_directory: ~/project/packages/headless
+    steps:
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: NPM package tests
+          command: |
+            npm install
```
mindctrl, line 205: Do you have to run `npm install` again? Looks like it runs in `package-build` and is persisted to the workspace, and then `package-test` attaches to the same workspace.

packages/headless/package.json
```diff
@@ -76,6 +76,8 @@
     "next": "^10.0.3",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.1.2",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
```
nickcernis, line 80: Were these reintroduced so that the Jest tests for the `src/components/menu/Menu.test.tsx` component pass?

Adding them back as `devDependencies` here causes `lerna bootstrap` to fail at the `npm run build` step. I think that's why @claygriffiths removed them in https://github.com/wpengine/headless-framework/pull/54.

Not sure of the best fix here, though — perhaps we could remove these lines as well as the Menu test for now and test the menu as part of our Cypress tests instead?

nickcernis (dismissed):

.circleci/config.yml
```diff
@@ -175,6 +175,37 @@ jobs:
...
+    docker:
+      - image: cimg/node:lts
+    working_directory: ~/project/packages/headless
+    steps:
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: NPM package install and build
+          command: |
+            npm install
```
mindctrl, line 188: Thoughts on using `npm ci` here?

packages/headless/package.json
rfmeier: I think these tests were introduced pre-TypeScript integration in which the react framework packages were swapped out for `@types/react`.

Do the tests need to be rewritten for TypeScript?

.circleci/config.yml
markkelnar: I did. The test failed even when the workspace was persisted.  I think I'm going to try and find a node/npm command, if there is one, that does the package.json caching, for the performance boost.

packages/headless/package.json
markkelnar: Adding this did allow the tests to pass.  I don't know much about how package.json devDep work, but added them in and things were 🟢  so I went with it. 😁

.circleci/config.yml
markkelnar: Maybe there's something about the npm install that sets something up in the environment. #still-learning 
There is https://circleci.com/developer/orbs/orb/circleci/node.  circleci/node and takes care of caching https://circleci.com/developer/orbs/orb/circleci/node#commands-install-packages.  I'm gonna try that.

packages/headless/package.json
markkelnar: Just removed Menu.test.jsx and reverted these package json changes.  Will that be sufficient for now?

.circleci/config.yml
```diff
@@ -175,6 +176,39 @@ jobs:
...
 
+  package-build:
+    executor:
+      name: node/default
+      tag: lts
+    working_directory: ~/project/packages/headless
+    steps:
+      - attach_workspace:
+          at: ~/project
+      - node/install-packages
```
mindctrl, line 187: Nice 😎

packages/headless/package.json
mindctrl: I think it's ok and we can add it back in another issue.

mindctrl (approved):

packages/headless/package.json
nickcernis: I agree, thanks for taking care of that @markkelnar.

nickcernis (approved):

PR #112, opened 2021-02-02
BREAKING CHANGE: Adjust domain replacement logic to remove header / query parameter requirement.

mindctrl:
Previously the plugin required enabling the "Enable Post and Category URL rewrites" setting
and also passing either the `replace-domain` parameter to the WPGraphQL endpoint
or sending `HTTP_X_WP_HEADLESS` in the header.

Now it only requires enabling the plugin option. No other request parameters
are required.

Related: #99


nickcernis (approved):
Looks good. Thanks for taking this on!

PR #115, opened 2021-02-03
fix: ensure WP stylesheets start with WP URL

nickcernis:
While documenting the `WPHead` component I noticed that stylesheets hotlinked to the WP instance are being rendered with an 'undefined' URL like this:

```
<link href="undefined/wp-includes/css/dist/block-library/style.css" rel="stylesheet">
```

Which causes a console error in Chrome due to the bad URL:

```
head-manager.ts?0645:83 Resource interpreted as Stylesheet but transferred with MIME type text/html:
"http://localhost:3000/undefined/wp-includes/css/dist/block-library/style.css".
```

This PR fixes things by passing the WP URL to Next.js via WPGraphQL instead of depending on `process.env`, which is `undefined` when the `WPHead` component uses it. I'm open to alternative solutions.

```
<link href="http://dev.test/wp-includes/css/dist/block-library/style.css" rel="stylesheet">
```

## To test
1. Create a WP page with a slug of `about`.
2. `npm run dev` and visit http://localhost:3000/about

View source or inspect the `head`. You should see a `block-library/style.css` `link` element prepended with your WordPress URL.

### DIFF
--- a/packages/headless/src/api/queries/GENERAL_SETTINGS.ts
+++ b/packages/headless/src/api/queries/GENERAL_SETTINGS.ts
@@ -5,6 +5,7 @@ const GENERAL_SETTINGS = gql`
     generalSettings {
       title
       description
+      url
     }
   }
 `;
--- a/packages/headless/src/components/WPHead.tsx
+++ b/packages/headless/src/components/WPHead.tsx
@@ -1,11 +1,6 @@
 import React from 'react';
 import Head from 'next/head';
 import { usePost, useGeneralSettings } from '../api';
-import { trimTrailingSlash } from '../utils';
-
-const WP_URL = trimTrailingSlash(
-  process.env.NEXT_PUBLIC_WORDPRESS_URL || process.env.WORDPRESS_URL,
-);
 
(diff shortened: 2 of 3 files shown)


mindctrl:
> While documenting the WPHead component I noticed that stylesheets hotlinked to the WP instance are being rendered with an 'undefined' URL

Is this bug present in our current preview example? I don't see it here, so just want to understand how to reproduce.

nickcernis:
> Is this bug present in our current preview example? I don't see it here, so just want to understand how to reproduce.

I see it for our preview example in canary (I'm at f1c0bdd) when doing `npm run dev` from the root or from `examples/preview`.

mindctrl:
@nickcernis thanks. I'm still not able to reproduce on `f1c0bdd` or `canary`. Can confirm this fix still works though. Curious if anyone else can reproduce.

nickcernis:
Thanks for your test, @mindctrl. I'll leave this for now as a record of the glitch until we can confirm it's reproducible.

claygriffiths (approved):

mindctrl (approved):

PR #126, opened 2021-02-08
refactor(BH-893): pass templates to TemplateLoader and Data Fetchers rather than auto loading from theme dir

claygriffiths:
Currently, templates are loaded from the `theme` directory if it exists using a Webpack context.

After chatting with DevRel, this is not ideal as it does not provide enough flexibility if templates are to be pulled in from other NPM packages.

Additionally, by leveraging imports and `dynamic`, the initial bundle size is smaller thanks to code splitting.

---

_`getting-started` example before_

Screen Shot 2021-02-08 at 2 20 46 PM

_`getting-started` example after_

Screen Shot 2021-02-08 at 2 16 18 PM

---

### Todo/Q's

- [x] Should we rename `TemplateLoader` to `NextTemplateLoader` since it depends on `dynamic`?
- [x] Make sure things work if `templates` param is not passed to `<TemplateLoader />` or the Data Fetchers
- [x] Update docs


wjohnsto (commented):
`TemplateLoader` relies on Next for `dynamic` and `getNextPageInfo`. This could be corrected by allowing the `uri` and `dynamic` to be passed into this component as props, or we could have a TemplateLoader that does not use `next` and one that does. If `TemplateLoader` does take in `uri` and `dynamic`, perhaps `NextTemplateLoader` can just wrap `TemplateLoader` and provide the correct props.

packages/headless/src/components/TemplateLoader.tsx
```diff
@@ -22,11 +21,14 @@ export interface WPTemplates {
...
+  uriInfo,
+  dynamicLoader = React.lazy,
 }: {
+  uriInfo: UriInfo | undefined;
   templates: WPTemplates;
+  dynamicLoader: (loader: () => Promise<Template>) => React.ComponentType;
 }): JSX.Element | null {
-  const pageInfo = useNextUriInfo();
-  const Component = dynamic(() => resolveTemplate(pageInfo, templates));
+  const Component = dynamicLoader(() => resolveTemplate(uriInfo, templates));
```
wjohnsto, line 31: This could create an issue when `useUriInfo` initially returns `undefined`. Perhaps resolveTemplate should do nothing if `uriInfo` is undefined. It looks like currently it will return the index component.

claygriffiths: It looks like `useUriInfo` will return `undefined` only on the following conditional: https://github.com/wpengine/headless-framework/blob/canary/packages/headless/src/api/hooks.ts#L155-L159

Is the only time that it'd return undefined if there's a `uriPrefix` set and you're not viewing a page with the prefix?

claygriffiths: I wonder if throwing an exception in `resolveTemplate` if `pageInfo` is missing is a better option? 🤔

My thought there is pretty much everything depends on `pageInfo` as far as Data Fetchers go.

wjohnsto: Hmm, I would think it's possible for `useQuery` to return `undefined` data since it will be using `useEffect` and will have to return something initially and update state once the request is done. You might not hit this if using SSR/SSG and calling `useUriInfo` with the URI we use on the server side, but you would potentially hit it on the frontend.

wjohnsto (approved):

wjohnsto (approved):

nickcernis (approved):
Looks good!

There are some references to the old `theme/` folder in the docs  on `examples/getting-started/wp-templates/front-page.tsx` that need removing or swapping to the new `wp-templates/` dir.

Is it worth renaming `getting-started/scss/theme` to `getting-started/scss/wp-templates` and updating references so that the mental mapping there makes sense? (SCSS files in the current `scss/theme` folder are each scoped to the `getting-started/theme/[name].tsx` files.)

These could be separate PRs if needed, though.

claygriffiths:
@nickcernis Awesome, thanks for the review! Another PR works 😄

PR #130, opened 2021-02-09
fix: remove hash when querying GetUriInfo in useUriInfo()

claygriffiths:
Remove hash when querying for URI info via `nodeByUri` in `useUriInfo()` to prevent URLs such as `/page#testing` from returning with no content.

Note, we already use `getUrlPath()` in `getUriInfo()`.

### DIFF
--- a/packages/headless/src/api/hooks.ts
+++ b/packages/headless/src/api/hooks.ts
@@ -114,7 +114,9 @@ export function useUriInfo(
     );
   }
 
-  localUri = trimOriginFromUrl(localUri);
+  localUri = getUrlPath(localUri);
+  // eslint-disable-next-line no-param-reassign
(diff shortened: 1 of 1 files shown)


nickcernis (approved):
Confirmed visiting `/about#test` resolves the query to `/about` and jumps to fragment id.

jasonbahl:
nitpick, and maybe better as a new issue than review comment, but looks like we're double importing utils in this file.

https://github.com/wpengine/headless-framework/blob/6abb613dd6802109206c3e5523edacd6c3d304c9/packages/headless/src/api/hooks.ts#L5-L17

claygriffiths:
@jasonbahl Nice find! Surprised ESLint isn't barking about that.

PR #131, opened 2021-02-09
fix: prevent 500 error on archives

nickcernis:
Archives such as `/category/uncategorized` will no longer produce a 500 error. For [BH-864](https://wpengine.atlassian.net/browse/BH-864).

### To test
From the project root:
- `npm run clean` (if not using a clean clone)
- `npm run bootstrap`
- `npm run dev`

Visit http://localhost:3000/category/uncategorized. You should see a list of posts instead of a 500 error.

### Notes
- Posts are currently latest posts instead of category-specific posts. This will be fixed in a future PR and depends on ongoing work in the custom-queries branch.
- Will need merge conflicts resolving once https://github.com/wpengine/headless-framework/pull/122 is merged.


packages/headless/src/api/nextFetchFromWP.ts
```diff
@@ -40,23 +40,28 @@ export default async function nextFetchFromWP({
     ? await resolveTemplate(pageInfo as UriInfo, templates)
     : undefined;
 
+  const isLatestPostsFrontPage =
```
nickcernis, line 43: `isLatestPostsFrontPage` could be part of `UriInfo` if we think it has use outside of this file.

It would enable `pageInfo?.isLatestPostsFrontPage` the same way we can currently do `pageInfo?.isSingular`.

mindctrl (approved):
This fixes the errors and, as noted in the first comment, there will be another PR to query the proper posts after work in the `custom-queries` branch is finished.

PR #141, opened 2021-02-15
ci: BH-763 npm publish job

markkelnar:
Add npm package publish for headless.
Add a '--dry-run' step which runs when merged to main or canary branches.
Real publish is triggered on tag in the format 'package/headless/1.0.0'.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,6 +208,26 @@ jobs:
           command: |
             npm run test
 
+  package-publish:
+    executor:
+      name: node/default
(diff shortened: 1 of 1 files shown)


rfmeier (approved):

nickcernis (approved):
Looks good. Thanks, @markkelnar.

We could perhaps remove the dry run once we're happy this is working smoothly. (I don't know that it needs to run on every merge to main/canary forever, as we'll have some merges that don't touch `packages/headless`.)

PR #142, opened 2021-02-16
fix: only returning revalidate for SSG

wjohnsto:


nickcernis (approved):
Confirmed the `examples/preview` app resolves preview URLs correctly.

Thanks for the fix, @wjohnsto.

Just a heads-up that the preview app now differs from the docs at https://developers.wpengine.com/docs/headless/framework/docs/previews (for example, we're now importing `HeadlessProvider` from '@wpengine/headless/react' instead of `@wpengine/headless`). Suspect it's on your roadmap to align them but mentioning just in case.

PR #150, opened 2021-02-18
BH-871: Generate post type labels

mindctrl:
This is a first pass at generating labels for custom post types. I updated the example data in the tests to match the expected label array structure. We may need to adjust some of this as we start building out the UI, like where we get the singular/plural names from exactly, but this should get us very close to what we need.


nickcernis (approved):
Looks good!

I compared labels with https://developer.wordpress.org/reference/functions/get_post_type_labels/#description and this looks correct to me.

I agree with your choices for default values. Nice work!

PR #154, opened 2021-02-22
feat: Pagination component (BH-856) and getting-started example

nickcernis:
- Adds a Next/WPGraphQL-based Pagination component to display next/previous links.
- Demonstrates its use in the getting-started `category.tsx` template.

Screenshot 2021-02-22 at 14 21 07

## To test
1. In WordPress, make sure you have 21 or more posts in the uncategorized category. (For example, via [wp post generate](https://developer.wordpress.org/cli/commands/post/generate/).) This is so you can see posts across three pages.
2. `npm run bootstrap && npm run dev`
3. Visit http://localhost:3000/category/uncategorized.

You should see clickable next/previous page links.

## Notes
- Uses [cursor-based pagination as recommended by WPGraphQL](https://www.wpgraphql.com/2020/03/26/forward-and-backward-pagination-with-wpgraphql/). The downside is that sites migrating from WP will lose `/page/2` style pagination URLs, because WPGraphQL does not support/encourage this.
- I implemented this in a way that does not depend on client-side requests, so that users do not have to fill in `NEXT_PUBLIC_WORDPRESS_URL`, and so that paginated URLs such as `http://localhost:3000/category/uncategorized/after/YXJyYXljb25uZWN0aW9uOjU3MTY5` are more easily cacheable.


examples/getting-started/wp-templates/category.tsx
```diff
@@ -1,42 +1,90 @@
...
 import { useRouter } from 'next/router';
 import { GetStaticPropsContext } from 'next';
 import { Footer, Header, Posts } from '../components';
 
-export default function Category(): JSX.Element {
-  const router = useRouter();
-  const urlParts = router.asPath.split('/').filter(Boolean);
-  const category = urlParts[1];
-  const posts = usePosts({
+const POSTS_PER_PAGE = 10;
```
nickcernis, line 9: This could come from WP's reading settings via a custom query or a future `useReadingSettings()` hook, but I hard-coded it here for now to keep it more simple.

trevanhetzel:
Dang, that [cursor-based pagination article](https://www.wpgraphql.com/2020/03/26/forward-and-backward-pagination-with-wpgraphql/) is an excellent read!

examples/getting-started/wp-templates/category.tsx
```diff
@@ -1,42 +1,90 @@
...
+  category: string; // 'uncategorized'
+  keyword?: string; // 'before'
+  id?: string; // 'abc123'
+}
+
+/**
+ * Converts a string or array into category URL parts assuming the
+ * form /category/uncategorized/[before|after]/[abc123].
+ */
+const getCategoryUrlParts = (path: string | string[]): CategoryUrlParts => {
```
trevanhetzel, line 26: Is there anyway to bake `CategoryUrlParts`, `getCategoryUrlParts`, and `getQueryOptions` into the `Pagination` component itself? This is a lot of boilerplate. I would so love to just import `Pagination` from `@wpengine/headless` and literally use it with the framework just intrinsically knowing about the current cursor.

packages/headless/src/next/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: Props): JSX.Element | null {
+  if (!pageInfo.hasNextPage && !pageInfo.hasPreviousPage) {
+    return null;
+  }
+
+  let classes = 'pagination';
+  if (pageInfo.hasNextPage) classes += ' has-next';
+  if (pageInfo.hasPreviousPage) classes += ' has-previous';
+
+  return (
```
trevanhetzel, line 36: As it stands right now, the component seems to mostly be a presentational component, with all the business logic living in the example file. This is fine, because people will likely want to use their own presentational component eventually, but I wonder if the business logic then could also be a component so as not to clutter the page template?

examples/getting-started/wp-templates/category.tsx
nickcernis: Thanks for the feedback! I've refactored in https://github.com/wpengine/headless-framework/pull/154/commits/3b084b9bb8bec883510a6f2d6991562ecb014b46 to move the boilerplate out of the template. People don't need to know about pagination or cursors now.

packages/headless/src/next/Pagination.tsx
nickcernis: Added a `categoryOptions` helper in https://github.com/wpengine/headless-framework/pull/154/commits/3b084b9bb8bec883510a6f2d6991562ecb014b46.

nickcernis:
Refactored to move category option logic out of the template as suggested by @trevanhetzel.

This is ready for another look.

The ESLint CI failure seems does not seem related to this PR.

mindctrl (approved):
LGTM. Works as expected. Per Slack chat, we ran into some issues that appear to be WPGraphQL issues.

https://github.com/wp-graphql/wp-graphql/issues/1552

https://github.com/wp-graphql/wp-graphql/issues/1714

Same behavior is visible in the [WPGraphQL example](https://codesandbox.io/s/wpgraphql-apollo-forward-backward-pagination-example-spq53?from-embed).

Given all that, I think we can merge this without any workarounds and work on fixes upstream.

PR #155, opened 2021-02-22
feat: Point getting-started to a live demo URL for a faster new user experience (BH-895)

nickcernis:
- Points the getting-started example to https://headlessfw.wpengine.com by default, by setting that URL as a fallback in the framework if no `.env.local` (or env vars) exist.
- Makes creation of `.env.local` and setting `WORDPRESS_URL` and `WP_HEADLESS_SECRET` optional and the recommended step 2, instead of a compulsory step 1. 
- Outputs info in the console if env vars are not set.
- Updates readmes to prompt users to create their app first, then later connect it to their own WP instance. (See new [readme quickstart](https://github.com/wpengine/headless-framework/tree/live-demo#quick-start) and [getting started docs](https://github.com/wpengine/headless-framework/tree/live-demo/docs/getting-started).)

## To test
1. Rename the `examples/getting-started/.env.local` if it exists to prevent it from loading (e.g. `_.env.local`).
2. `npm run dev` from the repo root.

You should see notices like this in the console, but the site will still spin up:

```
getting-started: DEMO MODE: The WORDPRESS_URL environment variable is not set. Using demo site URL of https://headlessfw.wpengine.com. Set WORDPRESS_URL to your WordPress site URL and install the WPGraphQL plugin to see your own content.
getting-started: The WP_HEADLESS_SECRET environment variable is not set. Install the WP Engine Headless plugin and set WP_HEADLESS_SECRET to the “Secret Key” from Settings → Headless to enable post previews.
```

3. Visit http://localhost:3000. You should see demo content from https://headlessfw.wpengine.com/.

If you add a `.env.local` with env vars, you'll need to restart the `npm run dev` process (as documented).

## Notes
Will also need to update docs at https://developers.wpengine.com/docs/headless/guides/getting-started/create-app, but this could wait until the updated package/plugin are published.


packages/headless/src/api/apolloClient.ts
```diff
@@ -15,14 +15,16 @@ import { CookieOptions, getAccessToken } from '../auth';
...
 );
 
 if (!WP_URL) {
+  WP_URL = 'https://headlessfw.wpengine.com';
+
   if (isServerSide()) {
-    throw new Error(
-      'WORDPRESS_URL and NEXT_PUBLIC_WORDPRESS_URL environment variables are not set. Please set WORDPRESS_URL (or NEXT_PUBLIC_WORDPRESS_URL if you wish to also use client-side requests) to your WPGraphQL endpoint.',
+    console.warn(
+      'DEMO MODE: The WORDPRESS_URL environment variable is not set. Using demo site URL of https://headlessfw.wpengine.com. Set WORDPRESS_URL to your WordPress site URL and install the WPGraphQL plugin to see your own content.',
```
trevanhetzel, line 27: Love the intentionality here!!

examples/getting-started/components/Header.tsx
```diff
@@ -16,7 +16,8 @@ function Header({
   // TODO: accept a `menuItems` prop to receive menu items from WordPress.
   const menuItems = [
     { title: 'Home', href: '/' },
-    { title: 'About', href: '/about/' },
+    { title: 'About', href: '/about' },
+    { title: 'Posts', href: '/category/uncategorized' },
```
trevanhetzel, line 20: Niiiice, now the example *just works* out of the box 😄

mindctrl (approved):

PR #156, opened 2021-02-23
BH-870: Register content types and fields with WPGraphQL

mindctrl:


nickcernis (approved):
Nice work on this, and on getting GraphQL tests hooked up too!

PR #160, opened 2021-02-24
fix: improvements to the Posts getting-started component

nickcernis:
- Wraps “read more” anchor in a Next.js Link (props @trevanhetzel for spotting this)
- Removes `count` prop. We don't need to artificially filter posts now that we can pass a post count in the query (example in front-page.tsx).

### DIFF
--- a/examples/getting-started/components/Posts.tsx
+++ b/examples/getting-started/components/Posts.tsx
@@ -7,7 +7,6 @@ interface Props {
   posts: WPGraphQL.Post[] | undefined;
   intro?: string;
   id?: string;
-  count?: number;
   heading?: string;
   headingLevel?: HeadingProps['level'];
   postTitleLevel?: HeadingProps['level'];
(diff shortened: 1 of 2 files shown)


examples/getting-started/components/Posts.tsx
```diff
@@ -37,32 +33,35 @@ function Posts({
...
                   />
-                  <a
-                    href={post.uri}
-                    aria-label={`Read more about ${post.title}`}>
-                    {readMoreText}
-                  </a>
+                  <Link href={post.uri}>
+                    <a
+                      aria-label={`Read more about ${
+                        post.title || ' the post'
```
mindctrl, line 56:
```suggestion
                        post.title || 'the post'
```

^ Removes extra space before fallback text, since there's already a space after "Read more about".

mindctrl (approved):
Nice. One small space tweak, otherwise looks good to go.

examples/getting-started/components/Posts.tsx
nickcernis: Well spotted, thanks!

PR #184, opened 2021-03-26
deps: bump Next.js to 10.0.9

nickcernis:
Update to Next.js 10.0.9 to remove a transitive dependency (cacache) that depended on a vulnerable version of ssri (6.0.1) triggering dependabot alerts.

The e2e test failure here is a separate issue and can be fixed independently.


mindctrl (approved):
Works great.

Issue #198, opened 2021-05-06
Breakout `core` Package


matt-landers:
This issue is in response to @wjohnsto's comment on #194 

We need to break the package up with the first package being `core` which should include (but not limited to):
The core package includes:

- Authentication and authorization
  - OAuth handlers
  - JWT Auth
  - Preview support
- Data fetching
  - Retrieve WP data based on the URL scheme
  - Fetch custom queries
  - Compose queries from fragments so that components can let the framework know the data it requires
  - Provide a mechanism for mutations
- Gutenberg block processing
  - Communication between WP and the frontend
  - Shared styles
  - Configuration
  - Data changes

wjohnsto:
I have a working branch and package for this here: https://github.com/wpengine/headless-framework/tree/feat-core/packages/core

wjohnsto:
The initial work for broking out the package is done as of #215. There is also #214 which breaks down auth. We need to break down data fetching and Gutenberg block processing further.

PR #215, opened 2021-05-13
`core` broken out

wjohnsto:
This PR satisfies #198.

NOTE: There is no name for the framework yet, so we are going to have to use a temporary alias while we wait on naming.


packages/core/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@wpengine/headless",
```
blakewilson, line 2: Lerna is failing due to two packages (`packages/core/package.json`, `packages/headless/package.json`) with the same name.

Maybe for now we could go with something like this until we finalize a name?

```suggestion
  "name": "@wpengine/headless-core",
```

blakewilson (commented):
Only thing I see is the core `package.json` name. That will need to be updated in order for Lerna to run. I've got this set to `@wpengine/headless-core` in the react package for now until we come up with a final name.

wjohnsto:
> 
> 
> Only thing I see is the core `package.json` name. That will need to be updated in order for Lerna to run. I've got this set to `@wpengine/headless-core` in the react package for now until we come up with a final name.

I saw you renamed it so I figured I would let you include that as part of your PR for the react components.

blakewilson:
@wjohnsto Perfect! LGTM then.

blakewilson (approved):

PR #234, opened 2021-05-20
[fix]: Incorrect rest_url with site_url pointing to headless front-end

### LINKED ISSUE #233: [wpe-headless] Site URL conflict
Hey All, I just ran across a "bug" and I'm not sure who needs to know so thought I'd send it to both of you. So yesterday I changed the "Site Address (URL)" to point to my atlas front end. ~~this fixed all the "View" buttons for posts and pages.~~ It fixes an issue in wpgraphql I talked with Jason on the other day where the `nodeByUri` uses the `site_url`. 

HOWEVER, once this is changed Gutenberg is broken. For some reason Gutenberg uses the frontend url and it breaks the ability to create/update posts. There are some solutions @ https://github.com/WordPress/gutenberg/issues/1761 .

image

moonmeister:
I don't do a lot of php,  let me know if I can fix anything. 

# Related issues

- fixes: #233

### DIFF
--- a/plugins/wpe-headless/includes/rest/url.php
+++ b/plugins/wpe-headless/includes/rest/url.php
@@ -10,15 +10,10 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-
 function fix_rest_url($url) {
-    $new_url = str_replace(home_url(), site_url(), $url);
-    return $new_url;
+    return str_replace(home_url(), site_url(), $url);
 }
 
-
 add_filter( 'rest_url', 'fix_rest_url');
-
-
 
 ?>
--- a/plugins/wpe-headless/includes/rest/url.php
+++ b/plugins/wpe-headless/includes/rest/url.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * Fixes rest_url when configuring front-end URL.
+ *
+ * @package WPE_Headless
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+
+function fix_rest_url($url) {
+    $new_url = str_replace(home_url(), site_url(), $url);
+    return $new_url;
+}
+
+
+add_filter( 'rest_url', 'fix_rest_url');
+
+
+
+?>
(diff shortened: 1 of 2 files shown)


mindctrl:
Thanks for the PR.

This issue came up before when we were doing discovery around whether or not we needed a custom setting for the front-end site URL. We decided it needed additional consideration to ensure we don't break things unintentionally.

Screenshot from Jira [BH-744](https://wpengine.atlassian.net/browse/BH-744):
image

moonmeister:
What are you saying? This PR doesn't use `site_url` for your headless use case. It enables folks set a unique address in Site URL to fix their "View" buttons for posts and page, and wpgraphql `postByUri` queries.

moonmeister:
Thanks for the clarification. I went and read the linked Jira issue but didn't realize that. (p.s. I'm one of the new SAs at WPE). Given this is all canary work I'd say this is the perfect time to break things and find out what happens...but it's up to ya'll. I have this in my custom theme for headless now and will report if I run into any issues. I don't for see any as it seems to be "fixing" the rest_url...can't imagine how it would break anything...but very well might.

mindctrl:
> (p.s. I'm one of the new SAs at WPE)

Nice, and welcome! Let us know if you run into any issues with that code running full time. If not, I think we could circle back to this and take a deeper dive.

On another note, you mentioned changing the Site Address to fix "View" links and such. Did the plugin not rewrite links for you automatically? We have some settings under Settings > Headless for post and category URL rewrites, which will update those links to point to the Front-end site URL specified on that page. Or was there a conflict with WPGraphQL?

moonmeister:
good catch. I realized halfway through this fix I didn't have the plugin installed cause I was playing with Gatsby on the site I was messing with and didn't need `wpe-headless`.  You are correct, the "View" issue is non existent with wpe-headless installed.

wjohnsto:
Question here @moonmeister and @mindctrl, as I'm unfamiliar with these filters. This fix allows you to set your Site Address to your frontend (e.g. http://localhost:3000), which will change the View links for your posts/pages and also still allow them to be saved. Does it also change the preview URLs (i.e. is there a need for us to be rewriting the preview URLs using our own "Front-end site URL")?

moonmeister:
Yes, changing the front end url would allow that but it's redundant. Both the WPGatsby plugin and wpe-headless for use with atlas already rewrite the preview url based on their own config. The Gatsby plugin does not handle the "view" links though I know it's on their list of things to fix.


The only advantage this fix has is fixing a wpgraphql issue where when querying for data by url if the domain is appended it must match what's in the WPSite url config.


As was said, there's an argument to be made that all these plugins should just be using the site url option instead of their own custom option but to be more risk averse, they just created their own options, for better or worse. 

I hope that answers your questions.

PR #248, opened 2021-06-01
Create tests for `core` package

### LINKED ISSUE #246: Create tests for `core` package
If necessary adjust exports to be accessible, and modify `index.ts` files to export the relevant package exports.

wjohnsto:
closes #246


packages/core/src/auth/authorize.ts
```diff
@@ -2,7 +2,8 @@ import { headlessConfig } from '../config';
 import isString from 'lodash/isString';
 import { parseUrl } from '../utils';
 import { CookieOptions, getAccessToken } from './cookie';
-
+import fetch from 'isomorphic-fetch';
+fetch;
```
blakewilson, line 6: Was this intentional?

blakewilson (changes requested):

packages/core/src/auth/authorize.ts
wjohnsto: Nope!

blakewilson (approved):

Issue #232, opened 2021-05-20
[PROPOSAL] Data Fetching


wjohnsto:
**NOTE:** This is still a work in progress, I will add to it as more ideas come to mind

Questions unanswered:

1. Should we be making API calls without the explicit direction of the application?
2. How can we support a variety of URL schemes for things like pagination, etc?
    - Keep in mind that Next.js (and possible other SSG frameworks) [does not support query params](https://github.com/vercel/next.js/discussions/17269) when using SSG, so we need to make sure we allow for URLs that might not be WP conventions. We also might need the WP plugin to coordinate this.
3. How should we allow for components to declare fields needed from specific queries?
4. How should we allow for addons to provide data fetching capabilities for specific queries?
5. ...?

This is a reference to #198 which discusses:

> - Data fetching
>   - Retrieve WP data based on the URL scheme
>   - Fetch custom queries
>   - Compose queries from fragments so that components can let the framework know the data it requires
>   - Provide a mechanism for mutations

## Retrieving WP Data
Currently `@wpengine/headless` has the following services:

  - `getPosts`: get a list of posts
  - `getContentNode`: get an individual post or page
  - `getGeneralSettings`: get the WP site settings
  - `getUriInfo`: get information about a URL using WPGraphQL's `nodeByUri` query

#### `getContentNode`
Something that can get confusing is how `getContentNode` will work for posts _or_ pages. We probably want two functions, one for getting a post and another for getting a page. This will simplify our queries and reduce some confusion:

  - `getPost`: gets a single post
  - `getPage`: gets a single page

#### Categories
We should provide functionality around the WPGraphQL `categories` query:

  - `getCategories`: get a list of categories
  - `getCategory`: get a single category
  - `getPostsByCategory`: gets a list of posts for a specific category
  - ...?

#### Tags
We should provide functionality around the WPGraphQL `tags` query:

  - `getTags`: get a list of tags
  - `getTag`: get a single tag
  - `getPostsWithTags`: gets a list of posts that contain specified tags

#### Menus
We should provide functionality around the WPGraphQL `menus` query:

  - `getMenus`: get a list of menus
  - `getMenu`: get a single menu

Issue #225, opened 2021-05-18
Create tests for `react` package


blakewilson:
We need a suite of tests for the basic functionality of the `react` package. This may include:
* utils
* Hooks
* `HeadlessProvider`

PR #249, opened 2021-06-01
Tests: Create test for React package

blakewilson:
This PR relates to #225 

Created tests for:
* `useGeneralSettings`
* `useUriInfo`

Also modified the exports in `packages/react/src/index.ts` to only export modules that are necessary.

I've left out tests for data fetching related functionality as it will likely change once we've settled on #232


packages/react/test/hooks.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { act } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+import {
+  GENERAL_SETTINGS,
+  getPostsQuery,
+  GET_URI_INFO,
+  headlessConfig,
+} from '@wpengine/headless-core';
+import React, { PropsWithChildren } from 'react';
+import { useGeneralSettings, usePosts, useUriInfo } from '../src/hooks';
```
wjohnsto, line 12: Looks like `usePost` and `getPostsQuery` are not used, I'm getting a TS compile error for them.

blakewilson: nice catch 👍  Weird, TS was compiling fine for me. I've just pushed an update

wjohnsto (approved):

Issue #262, opened 2021-06-03
Move from Apollo to gqless


wjohnsto:
I came across this and think it might be useful for the framework to support a number of data fetching scenarios. Let's look into it and see how it works https://github.com/gqless/gqless

Pay specific attention to:

1. How does it work outside of React/Next?
2. How does it work with React/Next?
3. How well is SSG (and SSR/ISR) supported?
4. Is there opportunity for abstraction and additional functionality that `packages/core`, `packages/react`, and `packages/next` can provide specific to Headless WordPress?

Comment on this issue with any interesting findings. Try to include some sample code with your comments.

UPDATE: We're moving to gqless. Here is how it will work:

1. `core` will ship with a default implementation of the gqless client and WPGraphQL schema.
2. `core` will publish a function for creating the client that allows overwriting of the schema with a custom version (supports custom WPGraphQL queries)
3. `core` will also allow intercepting fetch calls to handle the request, apply headers, etc.
4. `react` will use `@gqless/react` and provide a similar level of configuration for the React gqless client as `core`
5. The example apps will provide a standard way of creating the client and providing it to the framework for use in any helper functions.

Branch for this here: https://github.com/wpengine/headless-framework/tree/feature-gqless

blakewilson:
Just got done testing a few use cases in Next with SSG/SSR. So far, this seems like it would be useful in our framework. Here are some of my initial findings:

Take a super simple example where you would like to fetch the first 6 posts for the home page (similar to what we are doing in our examples):

```tsx
import { PropsWithServerCache } from "@gqless/react";
import { GetStaticProps } from "next";
import Link from "next/link";
import { prepareReactRender, useHydrateCache, useQuery } from "../gqless";

type PageProps = PropsWithServerCache<{}>;

export default function HomePage({ cacheSnapshot }: PageProps) {
  useHydrateCache({
    cacheSnapshot,
    shouldRefetch: false,
  });

  const query = useQuery();

  const posts = query.posts({ first: 6 });

  return (
    <div>
      <h1>My Home Page</h1>

      <h2>My recent posts</h2>
      <ul>
        {posts?.nodes?.map((post) => (
          <li key={post?.id}>
            <Link href={`/posts/${post?.slug}`}>
              <a>{post?.title()}</a>
            </Link>
          </li>
        ))}
      </ul>
    </div>
  );
}

export const getStaticProps: GetStaticProps<PageProps> = async (context) => {
[...]

PR #264, opened 2021-06-09
Removing apollo and adding gqless

wjohnsto:
Addresses #262 in the `core` project. Once we push this we can make the necessary changes in `react` and then `next`.


blakewilson (approved):
LGTM

PR #282, opened 2021-06-15
feat: (#278) create react example

### LINKED ISSUE #278: Create basic example for React only
We need an example project that only uses React and `core`/`react`. It doesn't need to be as involved as the `next` getting started, and it can be all CSR, but be sure to include posts, pages, categories, and previews.

blakewilson:
This PR resolves #278.

This example includes:
* Pages
* Custom Pages
* Posts
* Previews
* Basic Express server (for auth)
* Category archives
* CSR

Additionally, this PR adds the `apiUrl` to config to customize the base path for the API. This value defaults to the `wpUrl`. This allows the user to specify where their `auth` routes are located, in the case they are not using a framework with built-in SSR. For example, a Create React App with a custom Express server that handles these requests at `api.mysite.com`.


examples/react/getting-started/src/components/Posts.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </Heading>
+        )}
+        {intro && <p className={styles.intro}>{intro}</p>}
+        <div className="posts">
+          {isLoading && <PostListLoader />}
+          {posts &&
+            posts.map((post) => (
+              <div
+                className={styles.single}
+                key={post.id}
```
wjohnsto, line 45: Should be `key={post.id ?? ''}` to avoid issues while post is being fetched and `id` is undefined.

wjohnsto (approved):

PR #334, opened 2021-07-01
feat: (#326) adding wrappers for gqless queries

### LINKED ISSUE #326: Add wrapper hooks in `next` for all the standard gqless hooks
Right now we provide a custom `useQuery` hook that ensures we use the correct client for every request. We need to do the same for the following hooks:

1. `useMutation`
2. `useSubscripotion`
3. `useLazyQuery`
4. `useTransactionQuery`
5. `usePaginatedQuery`

wjohnsto:
resolves #326

### DIFF
--- a/packages/next/src/client.ts
+++ b/packages/next/src/client.ts
@@ -135,8 +135,28 @@ export function getClient<
     return client;
   }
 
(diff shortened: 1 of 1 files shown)


blakewilson (approved):
LGTM

PR #349, opened 2021-07-16
chore: adding version sync script

wjohnsto:
This will make sure to update package versions and keep them in sync whenever you run the `minor` or `patch` scripts from root.


blakewilson (approved):
LGTM

PR #353, opened 2021-07-20
feat: (#351) customize graphql url

### LINKED ISSUE #351: WP/GraphQL endpoint
Faust asks for the WP endpoint then adds the `/graphql` path. But I'm wanting to proxy the endpoint through a CDN like GraphCDN(https://my-app.graphcdn.app) which doesn't have the `/graphql` endpoint and I'm not sure how else that WP url is used. 

I'm recommending we allow for this use case by either not assuming the `/graphql` endpoint or allow a specific override for the graphql endpoint.

blakewilson:
This PR resolves #351


packages/core/src/config/config.ts
```diff
@@ -4,6 +4,8 @@ import trimEnd from 'lodash/trimEnd';
 import extend from 'lodash/extend';
 import isObject from 'lodash/isObject';
 import type { RequestContext } from '../api';
+import validUrl from 'valid-url';
```
moonmeister, line 7: this is a total nit, so no worries either way. If you don't want to add a dep you can just use the native `URL` type to validate the URI. 

```js
try {
  const uri = new URL(suspect);
} catch (e) {
  // if the constructor errors int's an invalid url
}
```

examples/next/getting-started/config.js
```diff
@@ -0,0 +1,9 @@
+const { headlessConfig } = require('@faustjs/core');
```
moonmeister, line 1: Nit: Maybe this be `faust.config.js` for clarity?

moonmeister (approved):
Looks good. Everything I commented is preference, take them or leave them as you choose.

packages/core/src/config/config.ts
wjohnsto: This is what I am thinking. valid-url is a good library, but it seems like overkill for what we need to do.

We also have [parseUrl](https://github.com/wpengine/headless-framework/blob/canary/packages/core/src/utils/convert.ts#L71). We could created an assertion function in `utils/assert.ts` called `isUrl` ti indicate whether something is a URL or just a string/path.

wjohnsto (changes requested):

examples/next/getting-started/config.js
wjohnsto: I support this as well, I think this is initially what @blakewilson was going to call it.

packages/core/src/config/config.ts
blakewilson: Yeah good thinking here, I'll remove this dep

examples/react/getting-started/src/config/index.js
```diff
@@ -0,0 +1,11 @@
+require('dotenv').config();
```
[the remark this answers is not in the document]
blakewilson: @wjohnsto So since the Express server to enable previews is in JavaScript we would need to setup a transpiler. I figure for an example app, that may be too much boilerplate

moonmeister:
LGTM

wjohnsto (approved):

PR #355, opened 2021-07-20
feat: sveltekit example

moonmeister:
My initial cleanup of my Svelte Kit example to get added here. I think more work needs to be done on it. Maybe just make it visually similar/identical to the  next example? Here's what I know still needs work:

- [ ] Post/page previews (we should be able to handle these without the `/preview` prefix needed for next
- [ ] Post list pagination
- [ ] Featured category page with list/pagination
- [ ] Custom page
- [ ] central headless config for gqless cli and app?
- [ ] ...
what else would we like to see done?


wjohnsto:
I don't know that the styling necessarily needs to be similar to the next/getting-started, but I do think we should demonstrate:

- [x] List post page
- [x] Single post page
- [x] Custom page (about)
- [x] Dynamic page
- [ ] Post list pagination
- [ ] Featured category page with list/pagination

josephfusco:
Closing this as our new focus is shifting to Nextjs & React.

PR #360, opened 2021-07-21
Fixing preview auth to provide config

wjohnsto:


blakewilson (changes requested):
I think we want to remove the `packages/core/package-lock.json` right?

blakewilson (approved):
LGTM

PR #368, opened 2021-07-22
Update outdated WP doc links

jordanmaslyn:
Some links needed new URLs based on latest refactor. Some links pointed to things that no longer exist at all. All fixed in this PR!


blakewilson (approved):
Nice! LGTM

PR #373, opened 2021-07-22
Add some WP notices for users to configure permalinks if they are unset

jordanmaslyn:


plugins/wpe-headless/includes/settings/views/headless-settings.php
```diff
@@ -41,6 +59,13 @@
 							<p class="error-message"></p>
 						<?php endif; ?>
 					</section>
+					<?php if ( ! get_option( 'permalink_structure' ) ) : ?>
+					<section>
+						<h4><?php esc_html_e( 'Enable Permalinks', 'wpe-headless' ); ?></h4>
+						<p><?php esc_html_e( 'Headless is easier with pretty permalinks! If you are unsure of which option to choose, go with "Post Name".', 'wpe-headless' ); ?></p>
```
moonmeister, line 62: THe issue with"Post name" is it's not what the templates expect and it'd share a template with pages. The latter might be okay but this recommendation should probably align with our starters....maybe?

moonmeister (commented):
this is a good step

plugins/wpe-headless/includes/settings/views/headless-settings.php
jordanmaslyn: After discussion, @moonmeister and I believe this to be a non-issue.

plugins/wpe-headless/wpe-headless.php
```diff
@@ -55,7 +55,7 @@
  */
 function wpe_headless_activated_plugin( $plugin, $network_wide ) {
 	if ( ! defined( 'WP_CLI' ) && ! $network_wide && WPE_HEADLESS_PATH === $plugin ) {
-		wp_safe_redirect( esc_url( admin_url( 'options-general.php?page=wpe-headless-settings' ) ) );
+		wp_safe_redirect( esc_url( admin_url( 'options-general.php?page=wpe-headless-settings&new_activation=1' ) ) );
```
blakewilson, line 58: We'll need to use [`esc_url_raw`](https://developer.wordpress.org/reference/functions/esc_url_raw/) instead of [`esc_url`](https://developer.wordpress.org/reference/functions/esc_url/), as `esc_url` will convert the "&" to an HTML entity. It looks like `esc_url_raw` is the preferred method as well for redirects.

https://developer.wordpress.org/reference/functions/esc_url_raw/#more-information

blakewilson (changes requested):

plugins/wpe-headless/wpe-headless.php
jordanmaslyn: Nice catch! Change made.

blakewilson (approved):
LGTM

PR #392, opened 2021-08-10
fix: polylang url change conflict

### LINKED ISSUE #390: Bug: Breaks Polylang URLs
## The current behavior

After some recent updates polylang URLS are bing broken by the Headless  Plugin. I have 3 languages. English, German, and French. French and German URLs should be prefixed with their respective language code (e.g. `/fr/posts/my-post`). This is the case when the plugin is disabled but upon enabling the plugin the prefix is removed and there is no more language prefix. 

This did work and one point. If I had to guess it was working on 0.5.5 of the plugin, not sure when it broke exactlly. I'll do some testing to see if I can when it broke and what the change was. 

## Applicable Versions

- WordPress version: 5.8, 
- Headless plugin: 0.5.7

## The expected behavior

I'd expect the headless plugin not to break Polylang language prefixes.

moonmeister:
fix: make priority really low to happen last so any plugins modifying url happen first and the domain swap happens last.

fixes #390

### DIFF
--- a/plugins/wpe-headless/includes/replacement/callbacks.php
+++ b/plugins/wpe-headless/includes/replacement/callbacks.php
@@ -89,7 +89,7 @@ function wpe_headless_image_source_srcset_replacement( $sources ) {
 	return $sources;
 }
 
-add_filter( 'preview_post_link', 'wpe_headless_post_preview_link', 10, 2 );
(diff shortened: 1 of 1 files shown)


wjohnsto (approved):
LGTM

PR #409, opened 2021-08-20
Auth Enhancements

blakewilson:


packages/core/src/api/client/index.ts
```diff
@@ -35,9 +34,7 @@ function createQueryFetcher(
 ) {
   return async function (query, variables): Promise<any> {
     const url = getGqlUrl();
-    const token = getAccessToken({
-      request: context,
-    });
+    const token = getAccessToken();
```
wjohnsto, line 37: Something to think about here, should we always apply the token or potentially enable two clients from core: an `authClient` and `client`? Then in `@faust/next` we could also export an `authClient` and `client` that use the same interface, but you can determine which you need to use when forming your queries?

packages/core/src/auth/token.ts
```diff
@@ -0,0 +1,75 @@
+import { fetchToken } from './authorize';
```
wjohnsto, line 1: Do we need checks in here to make sure this is only done client-side? We don't want to accidentally set `accessToken` server-side.

packages/core/src/api/client/index.ts
blakewilson: I think we can get away with one client since we can leverage `isAuthenticated` from the `useAuth()` hook

wjohnsto: The only thing I'm thinking here is that we are always applying the access token to the Authorization header, which ends up with extra work on the WP end as well as a potential for authentication errors even when an authenticated request isn't necessary, right?

blakewilson: Yeah, that is a fair point. I suppose logically we shouldn't be adding an access token to a request that doesn't need it.

blakewilson: I think an `authClient` and `client` would work, I would just want to make sure that doesn't require use to duplicate any logic within the clients.

packages/core/src/auth/client/accessToken.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    return null;
+  }
+}
+
+/**
+ * The interval in which the access token is check if a new one
+ * needs to be fetched
+ */
+export const ACCESS_TOKEN_EXP_CHECK_INTERVAL = 15000;
```
blakewilson, line 108: I believe this was my naming originally, but maybe we should call this `ACCESS_TOKEN_EXP_CHECK_INTERVAL_MS` since all other constants are in seconds.

packages/core/src/auth/server/middleware.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (status > 299) {
+        res.statusCode = result.response.status;
+      } else {
+        res.statusCode = 401;
+      }
+
+      res.end(JSON.stringify(result.result));
+    }
+  } catch (e) {
+    log(e);
```
blakewilson, line 64: We may need to return a response here. Possibly something like:

```tsx
res.end(JSON.stringify({"error": "unauthorized"}));
```

wjohnsto: maybe:
```
    res.statusCode = 500;
    res.end(JSON.stringify({ error: 'Internal Server Error' }));
```

wjohnsto (approved):

PR #427, opened 2021-09-01
Fix: (#422) Use most recent post/page revision if exists

### LINKED ISSUE #422: Bug: Existing post/page previews do not pull the latest revision
Viewing previews of existing posts/pages do not pull the latest revision. Instead, they show the published post/page content

## Applicable Versions

- `@faustjs/core` version: 0.9.0
- `@faustjs/react` version: 0.9.0
- `@faustjs/next` version: 0.9.0
- WordPress version: N/A

## Steps To Reproduce

1. Go to an existing page/post in the WordPress admin
2. Make changes
3. View Preview

## The current behavior

Upon clicking preview on an existing post or page, the user is given the most recently published content

## The expected behavior

Upon clicking preview on an existing post or page, the user is then given the most recent changes, regardless of publish status

blakewilson:
This PR resolves #422.

We now check if any revisions exist for the post/page. If they do, that is the most recent preview data for an existing post/page. Otherwise, we display the data from the original post/page query (draft, or new post/page)

### DIFF
--- a/packages/react/src/client.ts
+++ b/packages/react/src/client.ts
@@ -16,6 +16,9 @@ import merge from 'lodash/merge';
 
 export interface Node {
   id?: string | null;
+}
+
+export interface WithRevisions {
   revisions: (arg0: { first: number }) => {
     edges?: { node?: Node }[];
   };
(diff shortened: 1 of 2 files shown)


packages/react/src/client.ts
```diff
@@ -16,6 +16,9 @@ import merge from 'lodash/merge';
 
 export interface Node {
   id?: string | null;
+  revisions: (arg0: { first: number }) => {
```
wjohnsto, line 19: I think we might want a `WithRevisions` interface or something similar to this:

```ts
export interface Node {
  id?: string | null;
}

export interface WithRevisions {
  revisions: (arg0: { first: number }) => {
    edges?: { node?: Node }[];
  };
}

export interface RequiredQuery {
  posts: (args?: {
    where?: {
      categoryId?: number;
      categoryName?: string;
    };
  }) => unknown;
  post: (args: { id: string; idType?: PostIdType }) => (Node & WithRevisions) | null | undefined;
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  pages: (args?: any) => unknown;
  page: (args: { id: string; idType?: PageIdType }) => (Node & WithRevisions) | null | undefined;
  category: (args: {
    id: string;
    idType?: CategoryIdType;
  }) => Node | null | undefined;
  generalSettings?: unknown;
}
```

Since `category` does not have `revisions`.

wjohnsto (changes requested):

wjohnsto (approved):

PR #464, opened 2021-09-14
Reference correct path for DEVELOPMENT.md

bdurette:
When the docs got reorganized, this link was broken.


blakewilson (approved):
Nice catch, thanks @bdurette

PR #426, opened 2021-08-31
Preview Overhaul

blakewilson:
This PR makes previews more consistent by using `redirects()` in the `next.config.js` to redirect all requests with a `preview` query param to the `/preview` page.

The `withFaust` function was introduced to enable users to merge their existing next config with the defaults set in Faust.

Additionally, we are now adding the preview link to the WP Admin edit post page via `wp_localize_script` so we can get the correct post preview link, and avoid the bug in new draft post/pages in https://github.com/WordPress/gutenberg/issues/13998

Resolves #370 
Resolves #419

PR #478, opened 2021-09-16
Fixed preview URL for published posts.

torounit:
If the permalink structure is only origin, such as the front-end URL http://localhost:3000, the '/' will be stripped from the path of the preview URL.

It will look like `http://localhost:3000/postshello-world/?preview=true&p=1`.

Fixed it so that it is `http://localhost:3000/posts/hello-world/?preview=true&p=1`.


apmatthews:
Thanks for calling this out @torounit! I was able to reproduce the issue with the following steps:

1. Set your permalink settings to "Custom Structure: `/posts/%postname%/`"
2. Set your Front-end site URL to `http://localhost:3000`
3. Edit a published post and look at the preview URL.
4. See that the preview URL is something like `http://localhost:3000/postshello-world/?preview=true&p=1`, where all but the final slash in the URL path have been removed.

What you've done here seems to fix the problem, but I think there is an opportunity to clean things up a little further. Much of the path parsing that is being done in the preview link filter is no longer necessary since front-end previews now operate off of query args (#426) instead of the `/preview/` segment of the URL path. Without the need to inject `/preview/` in the path, [these lines](https://github.com/wpengine/faustjs/blob/c7fa179ea059cbecf68c93e2b1be591049302092/plugins/wpe-headless/includes/replacement/callbacks.php#L117-L123) and [these](https://github.com/wpengine/faustjs/blob/c7fa179ea059cbecf68c93e2b1be591049302092/plugins/wpe-headless/includes/replacement/callbacks.php#L143-L145) should be safe to remove. Would you mind trying this approach?

For future reference, please take a look at [CONTRIBUTING.md](https://github.com/wpengine/faustjs/blob/canary/CONTRIBUTING.md#instructions-for-contributing-code) if you haven't already. Things like including tests, using semantic commit messages, and filing issues before starting a PR go a long way towards helping us get your work reviewed and merged 🙂!

torounit:
@apmatthews removed the lines. it works!

natedunn:
@apmatthews Can we get this looked at again. Really would love to see this fixed and don't want @torounit's effort to go to waste. Thank you.

apmatthews (approved):
Sorry for the shamefully long wait on this. All looks good! Thanks again for the fix @torounit and thanks @natedunn for the nudge.

apmatthews:
Failing e2e getting started example test is, once again, due to a slash in the branch name. Not affected by this PR.

Issue #490, opened 2021-09-20
feat: Upgrade GQty to 2.0.0


blakewilson:
A new version of [`GQty`](https://github.com/gqty-dev/gqty) was released with official support for Unions. We should see what it takes to upgrade and do so in the packages and examples

blakewilson:
Closed via #490

PR #491, opened 2021-09-20
feat: (#490) use gqty v2

blakewilson:


wjohnsto:
We want to make sure this works with everything we have, I don't think it will be anything breaking. Also take a look at [this issue](https://github.com/gqless/gqless/issues/215#issuecomment-924371040) and do some testing around it as it. We may want to provide some documentation around how to do Union queries given that they are used heavily in WPGraphQL.

packages/core/package.json
```diff
@@ -44,34 +44,34 @@
...
     "deepmerge": "^4.2.2",
-    "gqty": "^1.1.3",
+    "gqty": "^2.0.0",
     "isomorphic-fetch": "^3.0.0",
     "lodash": "^4.17.21",
     "cookie": "^0.4.1"
   },
   "devDependencies": {
-    "@gqty/cli": "^1.1.4",
+    "@gqty/cli": "^2.0.0",
```
PabloSzx, line 53: 2.0.1 now :eyes:

PabloSzx:
gqty/cli was updated to 2.0.1 too, and a new generation is needed

wjohnsto (approved):

PR #502, opened 2021-09-24
Update custom-hooks.mdx

jacobwise:
Changing PageIdType to URI in example since SLUG doesn't exist yet.
Fixing usePage typo


josephfusco:
Hey @jacobwise thanks for the fix!

I went ahead and opened up https://github.com/wpengine/faustjs/pull/872 to resolve the merge conflicts as that file no longer exists.

PR #512, opened 2021-09-27
Use `.env.local` instead of `.env`

### LINKED ISSUE #447: use `.env.local` opposed to `.env` in examples

blakewilson:
This PR allows users of the `next/getting-started` example to use any naming of their environment files with the help of [`dotenv-flow`](https://www.npmjs.com/package/dotenv-flow).

Also changed the example to use `.env.local` by default instead of `.env`, as this is what Next.js prefers.

Resolves #447

### DIFF
--- a/.github/workflows/e2e-next-example.yml
+++ b/.github/workflows/e2e-next-example.yml
@@ -47,7 +47,7 @@ jobs:
     - name: copy env
       working-directory: e2e-app
       run: |
-        cp .env.sample .env
+        cp .env.local.sample .env.local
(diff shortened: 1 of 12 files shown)


wjohnsto (changes requested):
Should be good to go, but you probably want to rename the `.env.sample` file to `.env.local.sample`

blakewilson:
@wjohnsto Just realized the `.env.local.sample` was being git ignored. Fixed! Nice catch.

wjohnsto (approved):

PR #520, opened 2021-09-29
Feat isr timing

### LINKED ISSUE #516: Change the default ISR revalidate to 15 minutes
Right now ISR is set to 1 second, which seems a bit overzealous. It can also lead to problems when combined with `next/link`, which pre-fetches by default.

We should also create some documentation around this, and mention `next/link` refetching, and how you could configure this on a per-page basis.

wjohnsto:
Closes #516 

If we're good with this implementation then let's merge, this is just my thoughts on how it could work. We can adjust this PR/branch if we discover a better way to do it.


apmatthews:
I like this approach. My only feedback is that it's a little confusing to see both the `next/NextConfig` generated by `withFaust` and another `@faustjs/next/NextConfig` that's unrelated to the former. One option could be to rename `headlessConfig()`,  `@faustjs/core/HeadlessConfig` and `@faustjs/next/NextConfig` to `faustConfig()`, `@faustjs/core/FaustConfig` and `@faustjs/next/FaustConfig`, but we could always do that as a follow up.

It doesn't look like there would be any breaking changes if we merge as-is, but we should make a dev note about importing `headlessConfig` from `@faustjs/next` instead of `@faustjs/core` for Next.js projects. We'll want to make that change in the getting-started example as well.

wjohnsto:
@blakewilson and I were actually talking about this already. I think we do want to rename the `headlessConfig` because it doesn't make a lot of sense with the rename of the framework to faust. It would not be a huge breaking change, but it is worth considering.

blakewilson:
This looks nice. A few thoughts:

* Will there be an instances of conflicts when using `headlessConfig` when imported from both `@faustjs/core` and `@faustjs/next`? 
* If so ^, maybe now would be a good point to rename `headlessConfig` to `faustConfig`, as we would need to instruct users to update their `faust.config.js` to pull in the correct `headlessConfig` import

Overall, this config setup does have a nice separation of concerns. It will be convenient to have a solid place for Next.js specific config.

wjohnsto:
You should only need to import one or the other, since the headlessConfig in `@faustjs/next` calls through to `@faustjs/core`, however we might want to rename `@faustjs/core/headlessConfig` to `@faustjs/core/coreConfig` and `@faustjs/next/headlessConfig` to `@faustjs/next/nextConfig`.

We should also rename the interfaces. Maybe ~HeadlessConfig~ -> FaustConfig and ~NextConfig~ -> FaustNextConfig? (given that as @apmatthews points out `NextConfig` is a named interface that comes from `next`.

blakewilson:
@wjohnsto I think we should rename the interface from `headlessConfig` -> `faustConfig`. I'm not sure if we necessarily need to have `faustNextConfig`, as we would be doing the following import:

```tsx
import { faustConfig } from '@faustjs/next'
```

In this situation, importing `faustConfig` from `@faustjs/next` I would think that `faustConfig` was going to be Next specific anyway since it's coming from the `next` package.

However, I don't really have a hard opinion on that. I'm good either way.

wjohnsto:
> 
> 
> @wjohnsto I think we should rename the interface from `headlessConfig` -> `faustConfig`. I'm not sure if we necessarily need to have `faustNextConfig`, as we would be doing the following import:
> 
> ```tsx
> import { faustConfig } from '@faustjs/next'
> ```
> 
> In this situation, importing `faustConfig` from `@faustjs/next` I would think that `faustConfig` was going to be Next specific anyway since it's coming from the `next` package.
> 
> However, I don't really have a hard opinion on that. I'm good either way.

I'm wondering what happens if you bring a third package into the picture that also needs it's own config.

blakewilson:
Yeah that's a really good point. This may be the case when/if more functionality is added to '@faustjs/react'.

wjohnsto:
Part of me feels like maybe this is why we should have separate configurations per library. So in `faust.config.js` you would have to pull in both `@faustjs/core` and `@faustjs/next` in order to configure them.

blakewilson:
We should explore other packages and their approach i.e. Babel and their plugin system

apmatthews (approved):
LGTM!

blakewilson:
^ Agreed!

However, we may want to only merge this in once we are ready to release. Since this modifies the example starter, the example will be broken until we publish a release with the new config changes.

PR #521, opened 2021-09-30
Implement changesets

blakewilson:
This PR implements [changesets](https://github.com/atlassian/changesets).

> The changesets workflow is designed to help when people are making changes, all the way through to publishing. It lets contributors declare how their changes should be released, then we automate updating package versions, and changelogs, and publishing new versions of packages based on the provided information.

Using this workflow, we will rely on changesets to generate our changelogs and publish our releases using the `changesets/action` action. At this time I took out the existing changelog content, but we can backfill it if needed or just start fresh from this point on, since they were not up-to-date to begin with.

On push to `canary`, changesets will either create a PR, or add to an existing changeset PR. Once we are ready to release, we merge the PR into canary, from which point the changesets action will then release the packages.

Contributors can create changesets in their PRs by running `npm run changeset` from the monorepo root.


package.json
```diff
@@ -31,18 +31,23 @@
...
+    "version:status": "changeset status",
+    "release": "npm run build && changeset publish"
   },
   "devDependencies": {
     "@types/jest": "^27.0.1",
     "@wordpress/env": "^4.0.0",
     "rimraf": "^3.0.2"
+  },
+  "dependencies": {
+    "@changesets/cli": "^2.17.0"
```
wjohnsto, line 51: Is this needed in production?

wjohnsto (changes requested):
I don't know that we _need_ to make these changes, I'm just wondering if the dependency on @changesets/cli is needed for our CI or if it is just a dev dependency.

package.json
blakewilson: The changesets getting started does not specify any `--save-dev` flags:
https://github.com/atlassian/changesets/blob/main/docs/intro-to-using-changesets.md#add-the-changeset-tool

The docs could be wrong, but I do think we need it as a prod dependency since `devDependencies` wouldn't be installed at [this point in the release GitHub Action](https://github.com/wpengine/faustjs/blob/implement-changesets/.github/workflows/release-packages.yml#L40) due to `NODE_ENV=production`.

wjohnsto (approved):

PR #524, opened 2021-10-01
Updated the link to the guide document on the configuration page of wpe-headless plugin.

sawanoboly:
Hi,

The old link no longer exists, so I changed it to a link to the new document.


apmatthews:
Thanks @sawanoboly! Looks like this was fixed in #602.

PR #527, opened 2021-10-01
chore: include gqty as a monorepo dev dependency

apmatthews:
Fixes failed module resolution for `gqty` when running the `next/getting-started` example project from the root of the monorepo.


blakewilson (approved):
LGTM

Issue #492, opened 2021-09-22
Refine docs around setting up/fixing composer unit/E2E tests


blakewilson:
We should revisit the WPE Headless plugin unit and E2E tests to see if there is any room for improvement, in addition to refining the [development](https://github.com/wpengine/faustjs/blob/canary/DEVELOPMENT.md) docs around setting these tests up locally for users that want to contribute.

PR #530, opened 2021-10-04
e2e test improvements

apmatthews:
Related to #492, this lays some groundwork for further iteration on e2e testing and the plugin's CI/CD pipeline.

- Fixes existing e2e tests
- Updates and simplifies e2e testing instructions

### Testing
Follow the e2e testing instructions in [DEVELOPMENT.md ](https://github.com/wpengine/faustjs/blob/5399835f48c2f2bdca3c31bc9742de6e13eeb3e2/DEVELOPMENT.md#end-2-end-testing)and verify that the tests run successfully.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -109,7 +109,7 @@ Use [Codeception](https://codeception.com/) for running end-2-end tests in the b
 
 ### 2. Front-end Setup
 
-1. Create the following `.env.test.local` in `examples/next/getting-started`.
+1. Create the following `.env.test` in `examples/next/getting-started`.
 ```
 # Your WordPress site URL
 NEXT_PUBLIC_WORDPRESS_URL=http://localhost:8080
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -117,9 +117,9 @@ NEXT_PUBLIC_WORDPRESS_URL=http://localhost:8080
 # Plugin secret found in WordPress Settings->Headless
 WP_HEADLESS_SECRET=00000000-0000-0000-0000-000000000001
 ```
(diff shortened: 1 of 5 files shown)


DEVELOPMENT.md
```diff
@@ -107,23 +107,28 @@ Use [Codeception](https://codeception.com/) for running end-2-end tests in the b
    - In shell, run `chromedriver --version`. _Note: If you are using OS X, it may prevent this program from opening. Open "Security & Privacy" and allow chromedriver_.
    - Run `chromedriver --version` again. _Note: On OS X, you may be prompted for a final time, click "Open"_. When you can see the version, chromedriver is ready.
 
-### 2. Headless Site Setup
+### 2. Front-end Setup
 
-1. From within the headless site `examples/getting-started` copy `.env.test.sample` to `.env.test`.
-   - If you are using the provided Docker build, you will not need to adjust any variables in the `.env.testing` file; else, you can adjust the environment variables as needed.
+1. Create the following `.env.test.local` in `examples/next/getting-started`.
```
wjohnsto, line 112: Does it need to be `.env.test.local` or just `.env.test`?

wjohnsto (changes requested):

DEVELOPMENT.md
apmatthews: Either one works since we're using [dotenv-flow](https://www.npmjs.com/package/dotenv-flow). They generally recommend committing `.env.test` and not committing `.env.test.local`, which is why I picked the `.local` version. However, we already exclude all `.env*` files from git rather than just the `.local` ones, so I'd be fine with the shorter `.env.test`. I'll make that update.

wjohnsto (approved):

PR #531, opened 2021-10-04
Add option to disable redirect in `useAuth` hook when a user is unauthenticated

### LINKED ISSUE #529: Conditionally add a user profile menu or login in button in my site's header.
In my header component I've got code

```js
const { useAuth } = client.auth;
const { isLoading, isAuthenticated } = useAuth();
```

And
```jsx
{isAuthenticated ? (
        <UserProfileMenu />
      ) : (
        <Link href="/login" passHref>
          <Button ml={4} colorScheme="primary">
            Login
          </Button>
        </Link>
      )} 
```

When the user is not logged in - faustjs redirects to the login page. 

Ideally I'd like to way to conditionally show user profile menu or login in button in my site's header.

Thank you for the consideration.

Cheers,
Maurice

blakewilson:
Introduced `UseAuthOptions` on the `useAuth` hook that allows a user to specify the `shouldRedirect` property. This property defines if the `useAuth` hook should redirect to the appropriate url when unauthenticated.

```tsx

  const { isAuthenticated, authResult } = client.auth.useAuth({
    shouldRedirect: false,
  });

  // Undefined until the authenticated state is determined. Then true/false.
  console.log(isAuthenticated);

  /**
   * The result of the auth request. If the user is authenticated, the result will be "true".
   * Otherwise, the authResult will be an object with two keys: "redirect", and "login".
   * These values correlate to the "authType" config option for "redirect" and "local" respectively.
   *
   * @see https://faustjs.org/docs/next/reference/custom-hooks#useauth
   */
  console.log(authResult);

  return (
    <>
      <h1>My Authenticated Page</h1>
      {isAuthenticated === true && <p>My authenticated content</p>}

      {isAuthenticated === false && <p>You are not authenticated.</p>}
    </>
  );
}
```

Additionally, added the `Content-Type` response header in the `authorizeHandler` middleware.

Closes #529

### DIFF
--- a/packages/next/src/hooks/useAuth.ts
+++ b/packages/next/src/hooks/useAuth.ts
@@ -7,6 +7,17 @@ import trim from 'lodash/trim';
 import noop from 'lodash/noop';
 import type { NextClient } from '../client';
 
+export interface UseAuthOptions {
+  /**
+   * Specify if the useAuth hook should facilitate the redirect to the appropriate url.
+   *
+   * @default true
+   * @type {boolean}
+   * @memberof UseAuthOptions
+   */
+  shouldRedirect?: boolean;
+}
+
 export function create<
   Schema extends RequiredSchema,
   ObjectTypesNames extends string = never,
@@ -16,7 +27,8 @@ export function create<
     };
   } = never,
 >(): NextClient<Schema, ObjectTypesNames, ObjectTypes>['auth']['useAuth'] {
-  return () => {
+  return (useAuthOptions?: UseAuthOptions) => {
+    const { shouldRedirect = true } = useAuthOptions || {};
     const { authType, loginPagePath } = headlessConfig();
     const [{ isAuthenticated, isLoading, authResult }, setState] = useState<
       ReturnType<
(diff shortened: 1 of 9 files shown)


packages/next/src/hooks/useAuth.ts
```diff
@@ -16,7 +27,8 @@ export function create<
     };
   } = never,
 >(): NextClient<Schema, ObjectTypesNames, ObjectTypes>['auth']['useAuth'] {
-  return () => {
+  return (useAuthOptions?: UseAuthOptions) => {
+    const { shouldRedirect = true } = useAuthOptions || {};
```
wjohnsto, line 31: We should use `defaults` from `lodash` here instead of `useAuthOptions || {}`. While it should work this way, it could have unintended consequences.

wjohnsto (changes requested):

blakewilson:
@wjohnsto Thoughts on the naming of `shouldRedirect`? Does this seem clear enough?

wjohnsto:
> 
> 
> @wjohnsto Thoughts on the naming of `shouldRedirect`? Does this seem clear enough?

Naming is good

wjohnsto (approved):

Issue #622, opened 2021-10-27
Ensure Faust.js is compatible with Next 12


blakewilson:
With the relase of [Next 12](https://nextjs.org/blog/next-12), let's make sure that it is compatible with the latest version of Faust.js

PR #623, opened 2021-10-27
Add index to all core core directory imports

jordanmaslyn:
This should help with #622.

### DIFF
--- a/internal/website/sidebars.js
+++ b/internal/website/sidebars.js
@@ -13,7 +13,7 @@ module.exports = {
   docsSidebar: [
     {
       type: 'category',
-      label: 'Tutorial',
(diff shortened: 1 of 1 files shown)


blakewilson:
@jordanmaslyn Thanks for the PR! This is looking good.

Could you add indexes to the `@faustjs/next` package as well as increment `examples/next/getting-started` to `next@12.0.3`? We won't need to add indexes to `@faustjs/react` as it's already setup in this way.

Once that's done I can merge this in.

blistok:
@blakewilson @jordanmaslyn Not sure if you're planning to push this PR further, but I have addressed the feedback from here on #644.

PR #647, opened 2021-11-15
Deprecate `wpac/v1` api namespace in favor of `faustwp/v1`

apmatthews:
When work on the `wp-org` branch began, we decided to remove the `wpac/v1/authorize` endpoint in favor of `faustwp/v1/authorize`. In an effort to reduce the likelihood of errors due to a mismatch between a site's WordPress plugin version and it's front-end packages, this PR reintroduces the `wpac/v1/authorize` endpoint in a deprecated capacity.

It shouldn't matter whether a user updates their packages first or their WordPress plugin first -- their site should not break in either case. Achieving this cross-compatibility requires us to maintain deprecations that cover both cases.

While this PR should cover our immediate needs, I'm hoping to continue working on a more complete solution. Ideally, we could quickly and clearly warn developers about plugin/package incompatibilities (i.e. dashboard warning during FaustWP plugin update; some kind of npm post-update script; etc.). I have some concern that if we don't come up with a more complete solution, we could end up with a tangle of cross-compatibility deprecations that would be hard to reason about.

## Testing

- The existing e2e test for previews should cover the happy path (using the new auth endpoint).
- The core package will fallback to the deprecated auth endpoint when the new one returns a 404. This could happen when a user has updated `@faustjs/core` before updating the WordPress plugin. A unit test has been added to cover this behavior.
- Manual testing:
    1. Comment out the new [route registration in the FaustWP plugin](https://github.com/wpengine/faustjs/blob/dff41355deb1cfc21efef4da76338c039dd1eb28/plugins/faustwp/includes/rest/callbacks.php#L70-L78). (this will trigger both the package to execute its fallback behavior and the plugin to resolve the request using the deprecated route.)
    2. Ensure `FAUSTWP_SECRET` is set in your `.env.local`
    2. Verify that accessing a post or page preview still works.

### DIFF
--- a/packages/core/src/server/auth/token.ts
+++ b/packages/core/src/server/auth/token.ts
@@ -90,7 +90,9 @@ export class OAuth {
 
       if (response.status !== 404) {
         log(
-          'The FaustWP WordPress plugin is out of date. Please update to the latest version.',
+          'Authentication and post previews will soon be incompatible with ' +
+            'your version of the FaustWP plugin. Please update to the latest' +
+            ' version.',
         );
       }
     }
--- a/packages/core/src/server/auth/token.ts
+++ b/packages/core/src/server/auth/token.ts
@@ -4,6 +4,7 @@ import { config } from '../../config';
 import isNil from 'lodash/isNil';
 import isString from 'lodash/isString';
 import isNumber from 'lodash/isNumber';
+import { log } from '../../utils';
 
 export type OAuthTokenResponse =
   | OAuthTokens
@@ -87,10 +88,9 @@ export class OAuth {
         }),
       });
 
-      if (response.status !== 404 && process.env.NODE_ENV !== 'production') {
-        process.emitWarning(
+      if (response.status !== 404) {
+        log(
           'The FaustWP WordPress plugin is out of date. Please update to the latest version.',
-          'DeprecationWarning',
         );
       }
     }
--- a/packages/core/src/server/auth/token.ts
+++ b/packages/core/src/server/auth/token.ts
@@ -61,7 +61,7 @@ export class OAuth {
       );
     }
 
-    const response = await fetch(`${wpUrl}/wp-json/faustwp/v1/authorize`, {
+    let response = await fetch(`${wpUrl}/wp-json/faustwp/v1/authorize`, {
       headers: {
         'Content-Type': 'application/json',
         'x-faustwp-secret': apiClientSecret,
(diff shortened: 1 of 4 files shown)


packages/core/src/server/auth/token.ts
```diff
@@ -73,6 +73,28 @@ export class OAuth {
...
+          'x-wpe-headless-secret': apiClientSecret,
+        },
+        method: 'POST',
+        body: JSON.stringify({
+          code,
+          refreshToken: this.getRefreshToken(),
+        }),
+      });
+
+      if (response.status !== 404 && process.env.NODE_ENV !== 'production') {
```
apmatthews, line 90: Is this worthwhile? If so, is there a better environment variable we can check? I was thinking something that more explicitly asks for logging or debug info would be better, but don't know if such a thing exists.

blakewilson: we do have the [`disableLoggin`](https://github.com/wpengine/faustjs/blob/canary/packages/core/src/config/config.ts#L90) config option. This could be appropriate here.

plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -76,6 +76,17 @@ function register_rest_routes() {
 			'permission_callback' => __NAMESPACE__ . '\\rest_authorize_permission_callback',
 		)
 	);
+
+	// Deprecated. Faust.js packages now use faustwp/v1/authorize.
```
blakewilson, line 80: We should use the PHPDoc `@deprecated` tag here as well so we can easily find all deprecations.

apmatthews: I didn't use the tag initially since PHPDoc doesn't support comments and tags inside of functions, but there's nothing stopping us from doing it anyway. It looks nicer and is more consistent. Addressed in 510034c

moonmeister:
The WP Gatsby plugin has a built in compatibility API that means the client can confirm it's compatible. We don't have to copy this exactly but the client being able to check certain things could solve some issues.

blakewilson:
@moonmeister That would be an interesting approach. We'll have to look into that.

packages/core/src/server/auth/token.ts
```diff
@@ -73,6 +74,27 @@ export class OAuth {
...
+        method: 'POST',
+        body: JSON.stringify({
+          code,
+          refreshToken: this.getRefreshToken(),
+        }),
+      });
+
+      if (response.status !== 404) {
+        log(
+          'The FaustWP WordPress plugin is out of date. Please update to the latest version.',
```
blakewilson, line 93: I wonder if we should be more clear what is being deprecated that way users are not blindsided when we finally make the deprecation. 

Maybe something along the lines of "The `/wp-json/wpac/v1/authorize` endpoint will be deprecated soon. Please update to the latest version of the FaustWP plugin to be compatible."

Thoughts?

blakewilson (dismissed):

packages/core/src/server/auth/token.ts
apmatthews: I'm not sure how much detail we should give. I agree that a deprecation should usually warn you about what exactly is deprecated and how to fix it. However, I decided not to mention the endpoint in this case because it's not a public API. It's only consumed internally by the framework and isn't publicly documented. Our message isn't "we changed something that affects how you write your own code", but rather "we updated our own things, so you need to apply those updates to maintain compatibility." I think it's important for that distinction to be reflected in the message's wording.

Instead of mentioning specifics that could direct consumers attention towards their own code, we could just provide more detail on what features are affected. Maybe something like "Authentication and post previews will soon be incompatible with your version of the FaustWP plugin. Please update to the latest version."

blakewilson: Great point. I think that's perfect

blakewilson (approved):
LGTM 👍

PR #665, opened 2021-11-22
Deactivate WPE Headless on FaustWP activation

apmatthews:
## Description

While there shouldn't be any fatal conflicts between WPE Headless and FaustWP, running both plugins could lead to some unexpected behavior. We can try to prevent this by deactivating WPE Headless automatically during activation of FaustWP. Props @blakewilson for the idea!

## Testing

1. Run this PR's copy of FaustWP, but make sure the plugin is deactivated.
2. [Download the latest copy of WPE Headless](https://wp-product-info.wpesvc.net/v1/plugins/wpe-headless?download)
3. Install WPE Headless as a zip and activate.
4. Go to the Plugins page in the WordPress dashboard.
5. Click activate on FaustWP.
6. See that WPE Headless is deactivated.

### DIFF
--- a/plugins/faustwp/includes/utilities/callbacks.php
+++ b/plugins/faustwp/includes/utilities/callbacks.php
@@ -21,9 +21,10 @@ register_activation_hook( FAUSTWP_FILE, __NAMESPACE__ . '\\handle_activation' );
 /**
  * Callback for WordPress register_activation_hook() function.
  *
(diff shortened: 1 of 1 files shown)


blakewilson:
Works like a charm 👌 Nice @apmatthews

blakewilson (approved):

PR #644, opened 2021-11-14
Use fully qualified module imports for Next 12 compatibility

blistok:
## Related issues: 
- #622 
## Related PRs: 
- #623 (if merged this would supersede [#623](https://github.com/wpengine/faustjs/pull/623))

## Description:
This PR updates imports on all 3 packages:
- @faustjs/core
- @faustjs/next
- @faustjs/react

It also updates dependencies in `examples/next/getting-started` (Next is obviously bumped to major, others to minor/patch).

Next 12 will automatically extend `tsconfig.json` to include `incremental: true`, so I've added that as well.

All relative imports had to have full paths (with indexes - no directory resolution) [including extensions](https://github.com/nodejs/node/issues/30927).

For whatever reason I had to include extensions even to lodash imports (it wouldn't compile without it).

PR #667, opened 2021-11-30
Fix ESM import statements / Next 12 Support

blakewilson:
## Description

This PR addresses invalid ESM import statements by adding the file extension (and directory index when needed) that are required by ES Modules. As a result, this adds compatibility for Next.js 12.

File extensions are required for relative imports. They are also required for bare submodule specifiers (i.e. `import isUndefiend from 'lodash/isUndefined.js'`, when the `exports` field is not defined in `package.json`:

> Bare specifiers like 'some-package' or 'some-package/shuffle'. They can refer to the main entry point of a package by the package name, or a specific feature module within a package prefixed by the package name as per the examples respectively. Including the file extension is only necessary for packages without an "exports" field.

Below is some good reading regarding ES Modules' terminology and the requirement of file extensions:

- https://nodejs.org/api/esm.html#terminology
- https://nodejs.org/api/esm.html#mandatory-file-extensions

A small caveat in this PR is ESlint is being used to ensure imports have their appropriate file extension. However, linting errors are being thrown for our own bare submodule specifiers like `@faustjs/core/auth`. These are valid since we use the `exports` field in `package.json` like mentioned above, but ESlint doesn't know that.

This lack of compatibiiltiy between TypeScript/ES Modules will be addresses when [TypeScript 4.5 comes out of beta](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#esm-nodejs), and we could make the switch then to fix this ESlint band-aid.

## Related Issue(s):

- #622

## Testing

For some reason, ES Modules without their full path will sometimes work in the Faust.js monorepo. This makes for testing kind of a pain. Following the steps below will allow for proper manual testing:

### Validate the existing issue

1. In a seperate directory, create an `/examples/next/getting-started` app by following the [Getting Started guide](https://faustjs.org/docs/next/getting-started)
2. Update the `next` dependency in your newly created app to `^12.0.4`
3. `npm install`
4. Run `npm run dev`
5. Notice the ESM related errors

### Include the fix

1. `cd` into the Faust.js monorepo, and checkout the `esm-support` branch
2. Run `npm install` and `npm run build` to create the `dist` builds.
3. In your newly created app, `cd` into the `node_modules/@faustjs` directory
4. In each Faust.js package, remove the `dist` folder
5. Go back to the Faust.js monorepo, and copy the dist builds (`faustjs/packages/core/dist`, `faustjs/packages/react/dist`, `faustjs/packages/next/dist`) into your newly created app's node_modules (`node_modules/@faustjs/core/dist`, `node_modules/@faustjs/reactdist`, `node_modules/@faustjs/next/dist`)
6. Run `npm run dev`
7. Ensure pages, posts, and other functionality works as expected.

### Ensure the fix is backwards compatible with Next.js 11
1. Make a copy of your newly created apps' `node_modules/@faustjs` directory
2. Change the `next` dependency version back to `11.1.2`
3. `npm install`
4. Replace the newly created `node_modules/@faustjs` directory with the earlier copied version
5. Test to ensure latest fix works with Next 11

## Screenshots
Before fix:
Before fix

After fix:
After fix

## Related PRs

- #644
- #623

## Co-authors

- [jordanmaslyn](https://github.com/jordanmaslyn)
- [blistok](https://github.com/blistok)

Thanks to you both for your PRs!


apmatthews (approved):
LGTM!

theodesp (approved):

PR #668, opened 2021-11-30
Automate publishing changelog docs to website

apmatthews:
## Description

We have been using Changesets to automate GitHub releases and CHANGELOG.md updates, but these updates were not automatically included in the faustjs.org Release Notes page. This PR replaces the Release Notes page with a dedicated changelog page for each respective package. The content of each changelog page is imported directly from each package's CHANGELOG.md file, meaning that updates are automatically published to the website after the result of `npm run version` is merged (with the caveat that only pushes to `main` actually publish the site, so that step is still required).

I've also configured FaustWP to be (partially) versioned with Changesets, allowing us to publish it's changelog to the website alongside the JS packages. A script was added to `npm run version` that will handle updating the changelog in the plugin's `readme.txt` file to remain compliant with WordPress standards.

These changes are available for [preview on the staging site](https://hcixzyt38dn5ak04xxcqc36lf.js.wpenginepowered.com/docs/changelog/core).

There is still some polishing we can do to improve changelog formatting, but I've left that as a follow up.


internal/website/sidebars.js
```diff
@@ -177,9 +177,35 @@ module.exports = {
       id: 'gqty-intro',
     },
     {
-      type: 'doc',
-      label: 'Release Notes',
-      id: 'release-notes',
```
blakewilson, line 182: I wonder if we want to keep a page named `release-notes` with links to the appropriate changelogs and archive for SEO purposes/preserve the link in case anyone lands on it from Discord, etc.

blakewilson (approved):
This is awesome! I'm super excited about this. I left a comment, but I'll leave it up to you if you think it's worth while. Approved 👍

internal/website/sidebars.js
apmatthews: Yeah, that's a good call. I just moved the `/docs/changelog/archive` page back to `/docs/release-notes`. The content of that page already included the new location links at the top, so that worked out well. Organization of the site isn't impacted since the sidebar is defined separately from the page structure. I briefly considered setting up a redirect from `/docs/release-notes` to `/docs/changelog/archive` instead, but that seemed unnecessarily complicated.

[Preview](https://hcixzyt38dn5ak04xxcqc36lf.js.wpenginepowered.com/docs/release-notes)

blakewilson: Awesome! Looks great

PR #673, opened 2021-12-01
Suggestions for git workflow guidelines

apmatthews:
@blakewilson A diff seemed easier than review comments on #671.

The main thing I wanted to suggest was in regards to the staging workflow. I was thinking we would just merge the feature branch into `site-dev` locally and push to deploy. The changes would already be under review in the PR that targets `canary`, so we don't need the duplicate PR just to get things to the staging site. This could obviously leave staging in a potentially dirty or conflicted state, but I think that's a limitation of only being able to publish with branch pushes right now.

Everything else looks great!


blakewilson (approved):
LGTM!

PR #683, opened 2021-12-02
Fix plugin icon svg in chrome

apmatthews:
## Description

Figma has [an issue with linear gradient exports](https://forum.figma.com/t/svg-not-working-in-img-tag-or-as-css-background/9733/7) which causes SVGs in `img` tags to loose the background in Chrome.


blakewilson:
Here's the new raw SVG: https://raw.githubusercontent.com/wpengine/faustjs/eeeea1e6e850790872e9602ae0710dc048ff18e9/plugins/faustwp/.wordpress-org/icon.svg

Looks good in Chrome!

blakewilson (approved):

PR #685, opened 2021-12-02
Generate readme.txt changelog during versioning

apmatthews:
## Description

- Adds `scripts/buildPluginReadme.js`, which pulls the notes for the latest three versions in CHANGELOG.md into `readme.txt`.
- Runs the script automatically during versioning with `npm run version`.

## Testing

You can test the script by manually adding a new version section to the plugin's CHANGELOG.md, running `node scripts/buildPluginReadme.js`, and confirming that the changelog section of `readme.txt` is correctly updated.

### DIFF
--- a/scripts/buildPluginReadme.js
+++ b/scripts/buildPluginReadme.js
@@ -1,0 +1,71 @@
+/**
+ * Updates the FaustWP plugin's readme.txt changelog with the
+ * latest release found in the plugin's CHANGELOG.md file.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const util = require("util");
+const readFile = (fileName) => util.promisify(fs.readFile)(fileName, "utf8");
+const pluginPath = path.join(__dirname, '../plugins/faustwp');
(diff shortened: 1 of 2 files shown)


blakewilson (approved):
LGTM!

scripts/buildPluginReadme.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Updates the FaustWP plugin's readme.txt changelog with the
+ * latest 3 releases found in the plugin's CHANGELOG.md file.
+ */
+
+const fs = require("fs");
```
theodesp, line 6: Since Node v11 you can get promisified versions of `readFile` and `writeFile` so you can use them instead of the `util.promisify` and it becomes a little bit neater:

https://nodejs.org/api/fs.html#fspromisesreadfilepath-options

```
const fs = require('fs/promises');
const path = require("path");

const readFile = (filename) => fs.readFile(filename, {encoding: "utf8"});
const writeFile = fs.writeFile;
```

apmatthews: Addressed in #704

PR #691, opened 2021-12-03
misc(next): optimizations

moonmeister:
## Description

1.  Adding `sharp` for better production image processing.
2. Disabling compression on next servers since this is generally handled by platform/cdn as in Atlas. 
3. Changed some "Headless"  naming to "Faust.js" in the starter to better align with our branding. 

## Related Issue(s):

https://wpengine.atlassian.net/browse/NEI-27?atlOrigin=eyJpIjoiMTc2ZTU0ZjNkYmM2NDMyOGFlZDNlOGZmMTg5MDBmYzIiLCJwIjoiaiJ9

https://wpengine.atlassian.net/browse/NEI-28?atlOrigin=eyJpIjoiNGNjZGNiZDZlYWI0NDNkOGE1ZDIxNmM0ZDZhMWFhOGYiLCJwIjoiaiJ9

## Documentation Changes

Do we need any?


packages/next/src/config/withFaust.ts
```diff
@@ -52,5 +52,9 @@ export function withFaust(
   nextConfig.redirects = () =>
     createRedirects(existingRedirects, previewDestination);
 
+  if(nextConfig.compress == null) {
```
blakewilson, line 55: I'm not sure we would want to disable compression by default. It may lead to unintented consequences.

Atlas handles compression, but what about a platform that doesn't? How will that effect a user's build?

What is the downside of leaving `compress` in it's default state? Do the benefits outweight the possible unintended consequences?

blakewilson (changes requested):
@moonmeister 👋 ,

Thanks for the PR! I left a comment re:compression.

Also, two things:
* Looks like there were some linting errors. `npm run lint` will point you in the right direction there.
* Please delete `examples/next/getting-started/package-lock.json`, we don't track this since it's the example users clone via `npm create-next-app`

packages/next/src/config/withFaust.ts
moonmeister: It doesn't affect the build. This is wire compression, so `Accept-Encoding` is ignored and assets aren't compressed on the fly. If means their server will perform better albeit at the expense of transfer size. Most (All?) CDNs that proxy data handle compression for you. So Vercel, Atlas, Netlify, etc are all going to be unaffected or benefit from this change. Self hosting could be worse technically if they don't use a CDN. But I think we should choose defaults that are generally beneficial. The logic allows folks to override if desired. Maybe we document this somewhere to make it more obvious.

moonmeister: Leaving it as is just means servers potentially doing more work than needed if the edge/CDN could be handling compression.

blakewilson:
@moonmeister, apologies for the late follow up.

I like the sharp enhancements and wording updates. 

I think we should introduce:

```tsx
  if (nextConfig.compress == null) {
    nextConfig.compress = false;
  }
```

At a later time,  however. We're acting in a production ready stance right now, and introducing this will be a breaking chnage.

Perhaps we can add this when we go 1.0?

moonmeister:
> I think we should introduce:
> 
> ```tsx
>   if (nextConfig.compress == null) {
>     nextConfig.compress = false;
>   }
> ```
> 
> At a later time,  however. We're acting in a production ready stance right now, and introducing this will be a breaking chnage.
> 
> Perhaps we can add this when we go 1.0?


How is this a breaking change? If they've explicitly set it we honor the change. If not we disable it but nginx and CF will continue to handle compression. 

Happy to split this up just trying to understand your reasoning.

blakewilson:
Because `compression` is `true` by default in the Next.js config:
https://github.com/vercel/next.js/blob/canary/packages/next/server/config-shared.ts#L196

If you take the case of an existing Faust.js user who expects `compression` to be `true`, they update Faust.js, and now `compression` is `false`, that would be a breaking change for them.

apmatthews:
We have a ticket in the backlog to write some docs for deploying a Faust site to production. I wonder if we should just call out the compression setting there so users can make the right choice for their environment. If the environment happens to be Atlas, maybe we could provide an Atlas optimized config so they don't have to know details.

moonmeister:
> onder if we should just call out the compression setting there so users can make the right choice for their environment. If the environment happens to be Atlas, maybe we could provide an Atlas optimized config so they don't have to know details.

Was just thinking the same thing. I'll remove this for now and we can follow up with custom Atlas config later.

blakewilson (approved):
LGTM 👍

apmatthews:
Made a note about the compression discussion and linked things up in our backlog ticket for production docs https://wpengine.atlassian.net/browse/FST-7

Issue #619, opened 2021-10-27
Consider Interface for Custom Post Type Previews


blakewilson:
Currently, the `usePreview` hook only supports `post`/`page` post types. Let's look into supporting custom post types.

One challenge is that we do not have access to the TypeScript typings for custom post types at the `@faustjs` package level, since they live in `schema.generated.ts` at the app level.

The limitations of the current `usePreview` hook uses URL params to determine if the preview is for a `post` or a `page`, meaning you couldn't determine if the preview is for an alternative post type.

We could instead make a graphQL request using `contentNode` to return the proper post type, in addition to using the `$on` property to make a graphQL request to said content type for its preview data.

This would allow users to accurately determine the post type of the preview content, in addition to requesting the preview data for that post.

This would lead to an implementation in `preview.tsx` that looks like:

```tsx
import { PageComponent } from './[...pageUri]';
import { PostComponent } from './posts/[postSlug]';
import { TeamMemberComponent } from './team/[teamMemberSlug]';
import { client, Page, Post, TeamMember } from 'client';

export default function Preview() {
  const { usePreview } = client.auth;
  const { type, node } = usePreview();

  if (client.useIsLoading()) {
    return <p>loading...</p>;
  }

  switch (type) {
    // Default post
    case 'Post': {
      return <PostComponent post={node.$on.Post as Post} />;
    }

    // Page post type
    case 'Page': {
      return <PageComponent page={node.$on.Page as Page} />;
    }

    // Team member custom post type
    case 'TeamMember': {
      return <TeamMemberComponent teamMember={node.$on.TeamMember as TeamMember} />;
    }

    default: {
      return <div>Not found</div>;
    }
  }
}
```

blakewilson:
Closed via #710

PR #710, opened 2021-12-09
⭐ feat(previews): Create `usePreviewNode` hook

blakewilson:
## Description

Introduce the `usePreviewNode` hook for retrieving preview content for any post type. Deprecated `usePreview`.

## Related Issue(s):

- #619

## Testing

1. Checkout the `feat-usePreviewNode` PR
2. `npm install` from the monorepo root
3. `npm run build` from the monorepo root
4. Ensure current previews still function as expected (Posts, Pages)
4. Replace contents of `examples/next/getting-started/src/pages/preview.tsx` with:
    ```tsx
    import type { Page, Post } from 'client';

    export default function Preview() {
      const isLoading = client.useIsLoading();
      const { typeName, node } = client.auth.usePreviewNode();

      if (isLoading || node === undefined) {
        return <p>Loading...</p>;
      }

      if (node === null) {
        return <p>Post not found</p>;
      }

      switch (typeName) {
        case 'Page': {
          const page = node as Page;
          return (
            <>
              <h1>{page.title()}</h1>
              <div dangerouslySetInnerHTML={{ __html: page.content() }} />
            </>
          );
        }
        case 'Post': {
          const post = node as Post;
          return (
            <>
              <h1>{post.title()}</h1>
              <div dangerouslySetInnerHTML={{ __html: post.content() }} />
            </>
          );
        }
        // Add custom post types here as needed
        default: {
          throw new Error(`Unknown post type: ${typeName}`);
        }
      }
    }
    ```
5. Ensure previews will new preview code function as expected (Posts, Pages, drafts, existing, etc)
5. Create a new post type in your WordPress instance
6. Create a post in that post type
7. Go to `http://localhost:3000/preview?p=xxxx&preview=true&typeName=Post` where `xxxx` is your post ID
8. View preview content

## Documentation Changes

- Deprecated `usePreview` hook notice pushing users to use `usePreviewNode` instead


internal/website/docs/next/reference/hooks/usePreview.mdx
```diff
@@ -4,6 +4,12 @@ title: usePreview Hook
 description: The usePreview hook provides a way to get post preview data from WordPress
 ---
 
+:::caution Deprecation Notice
+
+`usePreview` has been deprecated in favor of [`usePreviewNode`](/docs/next/reference/hooks/usePreviewNode)
```
theodesp, line 9: Just wondering if its possible to log a deprecated usage warning whenever a client tries to use the `usePreview`?

internal/website/docs/next/reference/hooks/usePreviewNode.mdx
```diff
@@ -0,0 +1,82 @@
...
+    case 'Page': {
+      return <PageComponent page={node as Page} />;
+    }
+    case 'Post': {
+      return <PostComponent post={node as Post} />;
+    }
+    case 'Event': {
+      const event = node as Event;
+
+      return (
```
theodesp, line 71: I noticed that if we are using a switch statement it triggers two Graphql requests. One for getting the __typename and one for retrieving the title and content. For example I have a CUSTOM POST type called Movie:

```
 switch (postType) {
    case 'Page': {
      return <PageComponent page={node as Page}/>;
    }
    case 'Post': {
      return <PostComponent post={node as Post}/>;
    }
    case 'Movie': {
      const n = node as Movie;
      return <div>{n.isPreview}</div>
    }
    // Add additional custom post types here as needed
    default: {
      throw new Error(`Unspecified post type preview: ${postType}`);
    }
```

Then I see the requests in the console:
Screenshot 2021-12-09 at 16 41 47

Screenshot 2021-12-09 at 16 41 41

Is it possible to provide a way of avoid that? If not we should provide relevant documentation so that the users are aware of this behavior.

packages/next/src/gqty/hooks/usePreviewNode.ts
```diff
@@ -0,0 +1,89 @@
...
+      asPreview: true,
+    });
+
+    /**
+     * `contentNode` returns null if the post does not exist
+     * or if the preview has not been generated yet
+     *
+     * @link https://github.com/wp-graphql/wp-graphql/issues/2166
+     */
+    if (node === null) {
```
theodesp, line 69: You can use `(isNil(node))` as well.

internal/website/docs/next/reference/hooks/usePreviewNode.mdx
blakewilson: @theodesp Yes, this is correct.

Something I had in mind we could do is add the `postType` in the preview URL so we wouldn't have to do this second request:

https://github.com/wpengine/faustjs/blob/90ad8b7eaf1edebe11055b8114edb6dbeae8c218/plugins/faustwp/includes/replacement/callbacks.php#L139

packages/next/src/gqty/hooks/usePreviewNode.ts
blakewilson: Good catch! I'll made this change

internal/website/docs/next/reference/hooks/usePreview.mdx
blakewilson: This is a great idea. I'll add this.

blakewilson: On second thought, we may not be able to do this nicely. We could log a message in the `usePreview` hook, but it would log the message on every re-render and start to pollute the console.

package-lock.json
```diff
@@ -12700,12 +12700,13 @@
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg=="
     },
     "node_modules/sass": {
-      "version": "1.44.0",
-      "resolved": "https://registry.npmjs.org/sass/-/sass-1.44.0.tgz",
-      "integrity": "sha512-0hLREbHFXGQqls/K8X+koeP+ogFRPF4ZqetVB19b7Cst9Er8cOR0rc6RU7MaI4W1JmUShd1BPgPoeqmmgMMYFw==",
+      "version": "1.45.0",
```
theodesp, line 12703: I don't think this PR updates dependancies so we should revert those package-lock.json changes. Other than that it's good to go.

theodesp (approved):
LGTM.

internal/website/docs/next/reference/hooks/usePreviewNode.mdx
```diff
@@ -0,0 +1,87 @@
+---
+slug: /next/reference/hooks/usePreviewNode
+title: usePreviewNode Hook
+description: The usePreviewNode hook provides a way to get post preview data from WordPress for any post type
+---
+
+The `usePreviewNode` hook provides an abstraction around getting preview data for any post type. When calling the `usePreviewNode` hook on an appropriate preview page, the hook will ensure authorization, make a request to determine the post type, then make a subsequent request to retrieve the proper preview data. When using 0.7.1 of FaustWP or later, preview data will be retrieved with one request.
```
apmatthews, line 7: I believe the single query functionality would end up in FaustWP 0.7.2 or later since we're already on 0.7.1. That reminds me we need to add a changeset 🦋

package-lock.json
blakewilson: These are package lock changes that I believe were not versioned during the previous PR into `canary`, so I'm doing it now to make sure we're up to par.

PR #715, opened 2021-12-14
fix: correct admin url for plugin settings link

apmatthews:
## Description

Small update to the URL we use for the settings link on the plugin page. The former one got you to the right screen, but wasn't recognized and highlighted in the sidebar. We use `options-general.php?page=faustwp-settings` everywhere else, so this just keeps us consistent and restores the menu item's active state in the sidebar.


theodesp (approved):
LGTM

PR #724, opened 2022-01-04
A tiny readability adjustment to the docs website

aurooba:
This adjust the line length for better readability, so the docs are less of a headache to read and can be digested better. 

## Description

A tiny css change for the website, to adjust readability. :)

## Screenshots

Before:
Screen Shot 2022-01-04 at 12 14 22 PM

After:
Screen Shot 2022-01-04 at 12 14 15 PM


josephfusco (approved):

josephfusco:
@apmatthews It looks like SONAR_TOKEN is not available here causing the check to fail. Possibly related to https://github.com/wpengine/faustjs/pull/868?

apmatthews:
@josephfusco I think this might have something to do with external PRs not having access to GitHub secrets in workflows. Let me do some digging to see what the recommended approach is here.

PR #728, opened 2022-01-05
Properly invalidate refresh token upon expiration

blakewilson:
## Description

Currently refresh tokens that have been expired are not being properly invalidated and unset due to a bug. The logic is correct, however, it is in a conditional that it should not be in. This PR brings that logic outside of the conditional to properly invalidate the refresh token.

[Video demonstration of the bug](https://www.youtube.com/watch?v=ApqcCDfqaFg) (love this idea @jasonbahl)

## Testing

1. Clone Faust and create a page called `auth-content.tsx` with the following content:
    ```tsx
    
      const { isAuthenticated } = client.auth.useAuth();
      
      return <div>My Page</div>;
    }
    ```
2. On line 128 of `plugins/faustwp/includes/rest/callbacks.php`, set `$refresh_token_expiration` to some thing very short, like `5`.
3. Start the dev server running `npm run dev:next:getting-started` from the monorepo root
4. Go to `localhost:3000` and ensure you have no cookies saved for `localhost`
5. Go to `localhost:3000/auth-content`, notice that authorization happens, a refresh token is set, and authenticated content is now showing.
6. Wait 6 seconds for the refresh token to become expired
7. Refresh the page, and notice an infinite loop of authorization requests.
8. Now, checkout the PR
9. Follows steps 4-7, but now notice that the refresh token was properly unset, and a new authorization code was requested.

### DIFF
--- a/.changeset/cool-bikes-grab.md
+++ b/.changeset/cool-bikes-grab.md
@@ -1,0 +1,5 @@
+---
+'@faustjs/core': patch
+---
+
+Fixed a bug where expired refresh tokens were not being cleared from the browser cookie, possibly resulting in infinite loops during authentication, or an inability to request authenticated content.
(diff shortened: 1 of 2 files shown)


theodesp:
LGTM

theodesp (approved):

PR #735, opened 2022-01-12
🔨 refactor(examples): Enforce Node Engine

blakewilson:
## Description

Since Faust.js uses ES Modules, Node 12 will not work as intended. Specifying a Node version greater than 14 should help fix this.

### DIFF
--- a/examples/next/getting-started/package.json
+++ b/examples/next/getting-started/package.json
@@ -2,6 +2,9 @@
   "name": "next-headless-getting-started",
   "version": "0.1.0",
   "private": true,
+  "engines": {
+    "node": ">=14.0.0"
+  },
   "scripts": {
     "dev": "next dev",
     "build": "next build",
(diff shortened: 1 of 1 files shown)


examples/next/getting-started/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "next-headless-getting-started",
   "version": "0.1.0",
   "private": true,
+  "engines": {
+    "node": ">=14.0.0"
```
theodesp, line 6: Can we also add the minimum npm version?

```
"engines": {
    "node": "14.x",
    "npm": "6.x"
  },
```

apmatthews: @theodesp I saw you had [another issue in the backlog ](https://wpengine.atlassian.net/browse/FST-5) that was recommending `>=14.x` over `>=14.0.0`. Is that relevant here as well? Just want to make sure we're explicitly choosing one or the other. As far as I can tell they are equivalent 🤔

theodesp: Thats fine.

josephfusco (approved):
LGTM!

PR #738, opened 2022-01-20
Allow saving from WP's file editor

josephfusco:
## Description

Although the core hook [template_redirect](https://developer.wordpress.org/reference/hooks/template_redirect/) does not directly fire for wp-admin pages, it does fire for some admin facing actions like when attempting to save changes in WordPress' file editor.

The file editor attempts to perform a 'scrape' of the file by making a request to the front-end of the site with some query args.

```
http://wp.local/?wp_scrape_key=XXXXXXXXXX&wp_scrape_nonce=XXXXX
```

## Testing

1. Ensure “Disable WordPress theme admin pages” is disabled in FaustWP
2. Go to “functions.php” in active WordPress theme
3. Add a new function at the bottom of “functions.php”:
    ```php
    function test_save_file() {
      // Do nothing
    }
    add_action( 'wp_head', 'test_save_file' );
    ```
4. Save file.
5. Notice the following message. 
>File edited successfully.

### DIFF
--- a/plugins/faustwp/includes/deny-public-access/callbacks.php
+++ b/plugins/faustwp/includes/deny-public-access/callbacks.php
@@ -38,6 +38,11 @@ function deny_public_access() {
 	// Get the request uri with query params.
 	$request_uri = home_url( add_query_arg( null, null ) );
 
+	// Allow saving from file editor.
+	if ( false !== strpos( $request_uri, 'wp_scrape_key' ) ) {
+		return;
+	}
+
 	$response_code = 302;
 	$redirect_url  = str_replace( trailingslashit( get_home_url() ), $frontend_uri, $request_uri );
 


blakewilson (changes requested):
Nice! Could we modify the guard to also include checking for `wp_scrape_nonce`?

It looks like when both of these URL query params are set, WordPress will override the request to do the scrape:
https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/load.php#L1591

This could avoid someone intentionally setting a fake `wp_scrape_key` URL query param to get access to the WordPress page i.e. `https://my-wordpress-site.com/sample-page?wp_scrape_key=testing`

apmatthews (changes requested):
Nice work tracing this!

I think you're on the right track, but we should make sure we're validating that the request actually came from WordPress. If we only check for the presence of the string we end up creating a back door into any protected route that has `wp_scrape_key` in the query string. For example, `http://faustjs.local/my-protected-post/?bypass the-redirect-wp_scrape_key-success` would allow me to render the page in WordPress instead of redirecting to the front-end.

I think adding a function that does the same sort of [validation that WP does for a scrape request](https://github.com/WordPress/WordPress/blob/7f5d7f1b56087c3eb718da4bd81deb06e077bbbb/wp-includes/load.php#L1594-L1600) would be nice, but there may be other approaches.

apmatthews:
Ha, sorry for the double feedback! I didn't see @blakewilson's while I was writing mine.

josephfusco:
@blakewilson @apmatthews Thanks for the feedback! I've updated the PR with the requested changes.

apmatthews (approved):

Issue #739, opened 2022-01-21
Add the ability for some URLs not to be redirected to headless site


ghost:
There might be a case where a plugin would want to use native WordPress for a specific page or pages.

I propose adding a filter in callbacks.php of the WordPress plugin to allow for this.

anirudh7:
Why reinvent all the wheels, For example Membership feature, its hard to implement via headless, why not just use the membership options from core wordpress.

PR #740, opened 2022-01-21
feat[core]: (#739) allow plugins to filter out pages from public url redirect

cfg-fadica:
Offer the option to exclude specific pages from redirect should a developer have a need to do this.

## Description

The WordPress plugin offers the option to redirect all or none of the pages to the headless site URI. There may be cases where a developer would want some exceptions to this all or nothing approach. For instance if someone is using WooCommerce and wants to use the WooCommerce checkout page without having to rewrite all the logic in the headless site.

## Related Issue(s):
- #739 

## Testing

I added the following filter to a custom plugin I built after modifying callbacks.php to include this filter.

```PHP
add_filter( 'faustwp_exclude_from_public_redirect', function( $excluded ) {
  $excluded = array_merge( $excluded, [
    'checkout',
    'cart',
    'shop',
  ]);
  return $excluded;
}, 10, 1 );
```
This successfully prevented redirecting from https://mywordpresssite.com/shop/ to http://localhost:3000/shop. I tested other URLs such as https://mywordpresssite.com/alligator/ and that DID redirect correctly.

### DIFF
--- a/plugins/faustwp/includes/deny-public-access/callbacks.php
+++ b/plugins/faustwp/includes/deny-public-access/callbacks.php
@@ -32,6 +32,13 @@ function deny_public_access() {
 	if ( ! $frontend_uri ) {
 		return;
 	}
+	
+	// Allow for some URLs not to be redirected
+	$excluded_pages = apply_filters( 'faustwp_exclude_from_public_redirect', [] );
+
+	if( in_array( basename( add_query_arg( null, null ) ), $excluded_pages ) ) {
+		return;
+	}
 
 	$frontend_uri = trailingslashit( $frontend_uri );
 


theodesp:
Thank you @cfg-fadica. I think it's a really useful feature. We will review the proposal with the dev team and get back to you.

cfg-fadica:
While I had the code submitted previously in use on my site, I moved away from it, and now use a built in filter to accomplish this.

```PHP
<?php
/**
 * Exclude some pages from being redirected by Faust
 * @since 3.0.0
 */

 namespace CFG\Faust;

 use CFG\Interfaces\Hookable;

 class ExcludeRedirectPages implements Hookable {

    public function register_hooks() {
        add_filter( 'faustwp_get_setting', [ $this, 'exclude_pages' ], 10, 3 );
    }


    /**
     * Exclude a page from being redirected
     * @param mixed  $value   The setting value.
     * @param string $name    The setting name.
     * @param mixed  $default Optional setting value.
     */
    public function exclude_pages( $value, $name, $default ) {
        if( $name == 'enable_redirects' ) {
            
            /**
             * Disable redirects for certain WooCommerce pages.
             * 
             * @see /plugins/woocommerce/includes/wc-conditional-functions.php
             * @since 3.0.0
             */
            if( is_wc_endpoint_url() || is_checkout() || is_account_page() || is_privacy_policy() ) {
                return false;
            }

        }

        return $value;
    }

 }
```

The downside to this is that the $_SERVER variable is not available using this method.

josephfusco:
Thank you for the feedback and proposed changes!

Feel free to post your WordPress.org username (if you have one) and we'll be happy to get you added as a contributor on the [plugin page](https://wordpress.org/plugins/faustwp/).

Closing in favor of https://github.com/wpengine/faustjs/pull/1320.

PR #755, opened 2022-01-31
Deprecate template conditionals

josephfusco:
## Description

Now that we are using Next.js pages for routing, conditionalTags are no longer needed.

## Screenshots

Screen Shot 2022-01-31 at 9 52 05 AM

Screen Shot 2022-01-31 at 9 40 22 AM


blakewilson (approved):
LGTM 👍

PR #756, opened 2022-01-31
Sync `examples/next/getting-started` to Atlas Template

blakewilson:
This PR sets up a workflow to sync the example project (`examples/next/getting-started`) to the Atlas template for User Portal ([wpengine/faust-atlas-hello-world](https://github.com/wpengine/faust-atlas-hello-world)) when there are changes to `examples/next/getting-started` in the `main` branch of the `wpengine/faustjs` repo.

In order to allow the `wpengine/faustjs` repo to make commits in `wpengine/faust-atlas-hello-world`, we would need to either generate a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) (PAT), or use [deploy keys](https://docs.github.com/en/developers/overview/managing-deploy-keys), as the `GITHUB_TOKEN` env var that is loaded into GitHub Actions only has permissions for the repo it runs in.

PATs give access to all repos, and are linked to a specific user, so it is not ideal in this scenario. Instead, we'll use deploy keys. Deploy keys will allow us to make commits and pushes via an SSH key.

We'll have to generate an SSH key using the following command:
```bash
ssh-keygen -t ed25519  -C "Atlas Template Sync Key" -f path/to/keyfile
```

The SSH public key lives on the `faust-atlas-hello-world` repo as a deploy key, where the SSH private key will be stored in a secret in this repo, which we can access in our GitHub Actions workflow.

Todos before merging:
- [x] Setup deploy key on `faust-atlas-hello-world`
- [x] Setup SSH private key secret on `faustjs`

### DIFF
--- a/.github/workflows/sync-example-to-template.yml
+++ b/.github/workflows/sync-example-to-template.yml
@@ -11,6 +11,7 @@ jobs:
     env:
       CHECKOUT_DIR_NAME: ${{ github.event.repository.name }}
       SOURCE_REPO: ${{ github.event.repository.full_name }}
+      SOURCE_AFTER_SHA: ${{ github.event.after }}
       TARGET_REPO: wpengine/faust-atlas-hello-world
       TARGET_BRANCH: main
       SYNC_DIR: examples/next/getting-started # No trailing slash
@@ -52,7 +53,7 @@ jobs:
           git status
 
           # If there are changes, commit them
-          git diff-index --quiet HEAD || git commit -m "Automatic sync from $SOURCE_REPO"
+          git diff-index --quiet HEAD || git commit -m "Automatic sync from $SOURCE_REPO@$SOURCE_AFTER_SHA"
           
           # Push to the target branch
           git push origin $TARGET_BRANCH
--- a/.github/workflows/sync-example-to-template.yml
+++ b/.github/workflows/sync-example-to-template.yml
@@ -1,0 +1,58 @@
+name: Sync example project to Atlas template
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'examples/next/getting-started/**'
+jobs:
+  sync_example_project_to_atlas_template:
+    runs-on: ubuntu-latest
+    env:
+      CHECKOUT_DIR_NAME: ${{ github.event.repository.name }}
+      SOURCE_REPO: ${{ github.event.repository.full_name }}
+      TARGET_REPO: wpengine/faust-atlas-hello-world
+      TARGET_BRANCH: main
+      SYNC_DIR: examples/next/getting-started # No trailing slash
+      GIT_EMAIL: bot@wpengine.com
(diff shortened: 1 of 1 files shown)


theodesp:
@blakewilson thats great.

Did you tried this action just in case?
https://github.com/marketplace/actions/repo-file-sync-action

.github/workflows/sync-example-to-template.yml
```diff
@@ -0,0 +1,58 @@
...
+          # Add all the recent changes non-existent in the target
+          git add .
+
+          git status
+
+          # If there are changes, commit them
+          git diff-index --quiet HEAD || git commit -m "Automatic sync from $SOURCE_REPO"
+          
+          # Push to the target branch
+          git push origin $TARGET_BRANCH
```
theodesp, line 58: This might fail if someone pushes a new commit to this remote target branch that differs from the origin.

blakewilson: Good point. We could enable some kind of branch protection on `faust-atlas-hello-world`, or change line `58` to force push.

Since we have control of the repo, and the only commits will be coming from this workflow, I wonder what the probability of this happening is?

.github/workflows/sync-example-to-template.yml
```diff
@@ -0,0 +1,58 @@
...
+          git config user.email "$GIT_EMAIL"
+          git config user.name "$GIT_NAME"
+
+          # Add all the recent changes non-existent in the target
+          git add .
+
+          git status
+
+          # If there are changes, commit them
+          git diff-index --quiet HEAD || git commit -m "Automatic sync from $SOURCE_REPO"
```
apmatthews, line 55: Could we make the commit message reference the commit SHA in `wpengine/faustjs` that triggered the workflow? I think it could help us verify things are in sync at a glance.

Setting a var for [`github.event.after`](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push) and tacking it onto the `$SOURCE_REPO` to create the [full auto linked reference](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#commit-shas) might work. `$SOURCE_REPO@$AFTER_SHA`

blakewilson: Yeah good idea, just tried this out on a private repo, the full auto linked ref works great. Just pushed a commit for this

blakewilson: Yeah that is where I am leaning on the force push side of things. Say there is an edge case, we can always manually intervene. @theodesp @apmatthews Would you be comfortable running with this implementation, and then if any issues arise we can revisit?

apmatthews: I'm good with it ✅

josephfusco (approved):
Great solution @blakewilson!

PR #762, opened 2022-02-04
feat(useFaustContext.ts): Create and use `useFaustContext` in GQty queries

theodesp:

## Description

This PR:
* Adds a new hook `useCheckFaustContext` that will try to resolve the client passed through the `FaustProvider` value and throw an error if it's not provided.
* Fixes a wierd stackoverflow error when `FaustProvider` is missing.
* Forces the use of FaustProvider by calling the `useCheckFaustContext` in all core GQty hooks.
* Adds test case.

## Testing
* use the Next example folder and remove the <FaustProvider> from the `_app.tsx`. Observe the stack overflow.
* switch to this branch
* use the Next example folder and remove the <FaustProvider> from the `_app.tsx`. Observe the error message.

## Screenshots

**Before when using Faust without a context provider it throws a stack overflow error:**

Screenshot 2022-02-04 at 10 29 53

**After when using Faust without a context provider it throws an proper error message:**
Screenshot 2022-02-04 at 10 28 05


packages/next/src/gqty/hooks/useCheckFaustContext.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function useCheckFaustContext<T extends FaustContextType>(
+  reactContext: React.Context<T>,
+) {
+  const context = React.useContext(reactContext);
+  const client = context?.client;
+
+  if (client === undefined) {
+    throw new Error(
+      `Could not find "client" in the context. Wrap the root component in a <FaustProvider>`,
```
blakewilson, line 12: It might be helpful to put a link to the `<FaustProvider>` docs here:

```suggestion
      `Could not find "client" in the context. Wrap the root component in a <FaustProvider>. See: https://faustjs.org/docs/next/reference/faust-provider`,
```

blakewilson (commented):
Nice solution @theodesp!

PR #766, opened 2022-02-07
Remove plugin events callbacks

### LINKED ISSUE #588: Consider removing plugin's event callbacks
The WordPress plugin's `includes/events/callbacks.php` file contains a variety of error logging functions. It's only included when the `events_enabled` setting is true. Since that setting isn't surfaced in the UI, it's unlikely any of the event callbacks are being used. If possible, we should remove them.

apmatthews:
## Description

Removes event callbacks that, when enabled, print PHP `error_log` statements to track changes to rewrite rules and post/page statuses. These callbacks can only be enabled by a setting that doesn't actually exist in the plugin and removing them shouldn't functionally impact any existing sites.

## Related Issue(s):

Closes #588 

## Testing

Test suites should all still pass.


blakewilson:
Was there ever a time this option was available in the UI? Is there any reason to believe that a user has this option toggled on?

Cc @mindctrl @nickcernis

mindctrl:
@apmatthews @blakewilson I don't think these were ever an option in the UI. I think they were going to be used to notify systems to rebuild when new content is published, but nothing ever came of it. @rfmeier might remember more, I think he originally wrote it.

.changeset/cyan-timers-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'faustwp': minor
```
blakewilson, line 2: While we are removing functionality, it seems like it's functionality that was never exposed to the user. Thoughts on making this a `patch` release?

blakewilson (approved):
LGTM 👍

PR #767, opened 2022-02-08
[MERL-94] Docs Explain Headless Settings in detail now

theodesp:

## Description

* Adds new section in the docs for explaining the headless settings in detail
* Adds a new sidebar link to the FastWP plugin pointing to the new docs section

## Screenshots
Screenshot 2022-02-10 at 14 52 28


apmatthews (commented):
Nice work! Having this spelled out might help us just as much as it helps our users.

One thing that stood out to me is how we're using the terms `home_url`, `site_url` and `Front-end site URL`.  These can get a little confusing since `home_url` and `site_url` carry their own meaning in the context of WordPress.

https://developer.wordpress.org/reference/functions/home_url/
https://developer.wordpress.org/reference/functions/site_url/

These two settings get even more confusing in the context of a headless site, so for sake of clarity, maybe we avoid them altogether and just refer to them collectively as the WordPress URL? That would leave us with just two terms: **WordPress URL** and **Front-end site URL**. Using those terms consistently throughout these docs might help readers keep all these URLs straight 🙃

I'd also consider moving the documentation link closer to the settings checkboxes to make it easier to find (similar to the Secret Key input pictured below).

Screen Shot 2022-02-09 at 4 37 44 PM

blakewilson:
+1 to hyperlinking the appropriate section for each input

blakewilson (approved):

PR #770, opened 2022-02-10
With local auth example

blakewilson:
## Description

Add an example project to demonstrate [local auth strategy](https://faustjs.org/docs/next/guides/auth#local-based-authentication).

Lifted and updated from: https://github.com/blakewilson/faust-local-auth


theodesp:
That's great. Do we need to maintain this as close to the official example project though?

blakewilson:
Thanks! I don't think we'll need to maintain this one as close as our official example. I think keeping it in `examples/` works too. This structure mimicks what Next.js does:

https://github.com/vercel/next.js/tree/canary/examples

theodesp:
@blakewilson you can merge this now

theodesp (approved):

PR #774, opened 2022-02-11
Updates dependencies

apmatthews:
## Description

- Updates `@types/node` to latest version
- Updates `sharp` to latest version
- Removes outdated `@gqty/cli` as a devDependency from `@faustjs/core` since we don't generate schemas from there.
- Updates `@gqty/cli` in the example project.
- Includes a new schema from a blank WP site to reflect the updated output from `gqty generate`

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,9 +40,7 @@
     "release": "npm run build && changeset publish"
   },
   "devDependencies": {
+    "@changesets/cli": "^2.17.0",
     "rimraf": "^3.0.2"
-  },
-  "dependencies": {
-    "@changesets/cli": "^2.17.0"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -40,11 +40,7 @@
     "release": "npm run build && changeset publish"
   },
   "devDependencies": {
-    "@types/jest": "^27.0.2",
```
blakewilson, line 43: There was one point in time where our Jest types were not working from the package level, so we added `@types/jest` to the monorepo root here. It looks like this isn't the case anymore though. If it somehow crops up in the future we can just add this back.

PR #777, opened 2022-02-16
Create a guide for testing Faust.js apps with Jest

blakewilson:
## Description

Include a guide with contents/examples on how to test Faust.js apps with Jest.

### DIFF
--- a/internal/website/docs/next/guides/testing-with-jest.mdx
+++ b/internal/website/docs/next/guides/testing-with-jest.mdx
@@ -117,7 +117,8 @@ function Header({
               <Link href="https://github.com/wpengine/faustjs">
                 <a
                   className="button"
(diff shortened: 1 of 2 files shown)


apmatthews (commented):
Good stuff! Could we add one more line that wraps things up at the end after creating `Header.test.tsx`? Something like:

> Use `npm run test` to run the test and see that it passes.

Also, not sure if it's anything we can fix here, but I got this warning after running the test:

```
Jest did not exist one second after the test run has completed.
```
Screen Shot 2022-02-22 at 4 48 35 PM

blakewilson:
@apmatthews Yes, great idea! And the new access token refetch logic fixes that 😎

apmatthews (approved):

PR #782, opened 2022-02-18
Validate and sanitize FaustWP settings

apmatthews:
## Description

Uses a WordPress hook for sanitizing our settings option before it's written to the database. [Prior art in WordPress core](https://github.com/WordPress/WordPress/blob/2eaeae58ff0f4e7a5e31220830d6de060359322a/wp-includes/formatting.php#L4841-L4852) also suggests that this is a good place to validate settings and return errors when appropriate, so I've done that as well.

This _may_  impact some existing settings, but that impact should be limited to edge cases where we're dealing with bad values anyway. Here's a little more detail on what users can expect for each specific setting:

- Checkboxes and secret keys were well controlled in code before, so any existing values for those settings should already pass the new rules, unchanged.
- Menu locations are now run through `sanitize_text_field()`. [See the function docs](https://developer.wordpress.org/reference/functions/sanitize_text_field/) for details on what is stripped. This will only be applied when a user attempts to update or re-save existing settings.
- The front-end URL now explicitly requires `http://` or `https://`. Users that do not have a protocol on their front-end URL will have to add one next time they try to update any FaustWP settings. I think this is acceptable given that no-protocol URLs were implicitly unsupported and these users would already be experiencing bigger issues (such as previews not working) when neglecting to add one.

## Testing

Unit and e2e tests included. If you'd like to do any manual testing, the test cases have good sample data.

## Documentation Changes

Updated the `DEVELOPMENT.md` guide to use a valid UUID format for the e2e test environment so that it passes our new validation. I don't know of any other docs that need updates.


plugins/faustwp/tests/integration/settings/test-callbacks.php
```diff
@@ -0,0 +1,190 @@
...
+		'enable_rewrites' => '1',
+		'disable_theme' => '1',
+	];
+
+	public function setUp() {
+		parent::setUp();
+		update_option( $this->option, $this->init_settings );
+	}
+
+	public function test_the_sanitize_option_filter_is_registerd() {
```
theodesp, line 30:
```typo
test_the_sanitize_option_filter_is_registered
```

plugins/faustwp/tests/integration/settings/test-callbacks.php
```diff
@@ -0,0 +1,190 @@
...
+			$this->assertCount( 1, $errors );
+			$this->assertEquals( $errors[0]['code'], 'faustwp_invalid_frontend_uri' );
+			// Clear errors between test cases.
+			$wp_settings_errors = [];
+		}
+	}
+
+	public function test_sanitize_faustwp_settings_allows_valid_secret_key() {
+		global $wp_settings_errors;
+		// We have to clear this manually since all tests execute in a single "request"
```
theodesp, line 99: Maybe move this comment on the top of the file as it is duplicated across the test file.

apmatthews: Should be much cleaner in [5e79c55](https://github.com/wpengine/faustjs/pull/782/commits/5e79c55c3f9f2088def907ca9afe924389fdf1eb)

theodesp (approved):

PR #786, opened 2022-02-21
Explicitly refresh access token

blakewilson:
## Description

Currently, the access token is refreshed on a `setInterval`, globally polluting the module scope. This PR explicitly refreshes the access token by setting a timer to be invoked 60 seconds before the existing token expires.

Not only will this be more efficient, but will improve the testing experience as well as a timer is not being invoked globally when importing the `@faustjs` packages.

## Testing

Tests have been provided.

Manual testing:

1. Add the following before the `setRefreshTimer` statement of `setAccessTokenRefreshTimer()`:
   ```ts
   const expiresDate = new Date(
     new Date().getTime() + secondsUntilExpiration * 1000,
   ).toString();
   const refreshDate = new Date(
     new Date().getTime() + secondsUntilRefresh * 1000,
   ).toString();

   console.log(
     `Access token expires at ${expiresDate}, access token will be re-fetched at ${refreshDate}`,
   );
   ```
2. Go to an authenticated endpoint (previews for example).
3. Notice the console message and watch for the access token to be properly fetched at the specified time:
   Screen Shot 2022-02-21 at 9 08 15 AM

### DIFF
--- a/packages/core/src/auth/client/accessToken.ts
+++ b/packages/core/src/auth/client/accessToken.ts
@@ -31,9 +31,7 @@ export const TIME_UNTIL_REFRESH_BEFORE_TOKEN_EXPIRES = 60;
 /**
  * The setTimeout instance that refreshes the access token.
  */
-export let __REFRESH_TIMER: RefreshTimer = undefined;
-
-let accessToken: AccessToken | undefined;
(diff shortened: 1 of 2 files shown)


apmatthews (approved):
Much better!

josephfusco (approved):
LGTM!

PR #788, opened 2022-02-23
Fix previews when `trailingSlash` is enabled in Next.js config

apmatthews:
## Description

I wasn't able to find a single regex that works both when `trailingSlash` is enabled and when it's disabled, but we _can_ check the config and modify the regex appropriately. If `trailingSlash` is enabled, we expect the `/` at the end of the `/preview` `source` path. We also add it to the `destination` to avoid triggering another `trailingSlash` redirect.

Once that was corrected, I still had issues previewing. The `apiRouter` was returning a 404 on the `auth/token/` endpoint because of the trailing slash, hence the added `trimEnd()`.

## Testing

Jest tests were added. I'd like to cover this in an e2e test as well, but booting a totally separate front-end with a different next.config.js just to test for `trailingSlash` seemed excessive. Maybe one day we'll have a way to easily swap configs in our e2e environments, but for now I'd recommend testing this manually:

1. Run the stock example project `npm run dev`
2. Open a preview link from your WordPress instance
3. Verify that you're redirected to `/preview` and the preview is rendered
4. Stop the example project server and add the following to your `next.config.js` in `examples/next/getting-started`:

```js
const { withFaust } = require('@faustjs/next');

/**
 * @type {import('next').NextConfig}
 **/
module.exports = withFaust({
  trailingSlash: true,
});
```
5. Start the example project again `npm run dev`
6. Open a preview link from your WordPress instance
7. Verify that you're redirected to `/preview/` and the preview is rendered

### DIFF
--- a/.changeset/modern-llamas-shop.md
+++ b/.changeset/modern-llamas-shop.md
@@ -1,0 +1,6 @@
+---
+'@faustjs/core': patch
+'@faustjs/next': patch
+---
+
+Fixed previews when trailingSlash is enabled in Next.js config
--- a/packages/core/src/server/router/index.ts
+++ b/packages/core/src/server/router/index.ts
@@ -1,5 +1,6 @@
 import { IncomingMessage, ServerResponse } from 'http';
 import isUndefined from 'lodash/isUndefined.js';
+import trimEnd from 'lodash/trimEnd.js';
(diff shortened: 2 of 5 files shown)


theodesp:
It would be great if we refine and prioritise MERL-137 to have access to Integration tests for the future.

josephfusco (approved):
Working with trailing slash true & false 👍🏼

PR #790, opened 2022-02-25
handleSitemapRequests() Faust.js Middleware

blakewilson:
## Description

Create a Next.js middleware to handle WordPress sitemaps. The API is as follows:

```tsx
import { handleSitemapRequests } from '@faustjs/next';

handleSitemapRequests(middlewareReq, {
  // Your WordPress URL
  // The path to the root sitemap index
  sitemapIndexPath: '/wp-sitemap.xml',
  /**
   * Sitemap paths to ignore. Useful if you don't want to include
   * sitemaps for users, categories, etc.
   */
  sitemapPathsToIgnore: [
    '/wp-sitemap-posts-page-1.xml',
    '/wp-sitemap-posts-post-*', // allows wildcards as well to avoid multiple indices if necessary
  ],
  /**
   * List of Next.js pages to include in your sitemap.
   */
  pages: [
    {
      path: '/about',
      priority: 0.75,
      changefreq: 'monthly',
      lastmod: new Date().toISOString(),
    },
  ],
  /**
   * Replace WP urls with the headless frontend.
   */
  replaceUrls: true,
});

```

## Testing

Test cases have been provided in the PR.

### Manual Testing
1. View `src/pages/_middleware.ts`
2. Notice we are using `/wp-sitemap.xml` as our sitemap index. We're also excluding any sitemaps from the index that start with `/wp-sitemap-users-*`, and we are replacing urls from the WP site to the headless frontend
3. Compare the headless sitemap (http://localhost:3000/wp-sitemap.xml) that is handled by our api, to the WordPress sitemap (http://your-wp-site.com/wp-sitemap.xml)
4. Notice the authors sitemap is not added to the headless sitemap index, and we are also replacing URLs
5. Visit each headless sitemaps in the sitemap index and verify the URLs are correct with proper replacements
6. Add `/wp-sitemap-posts-page-1.xml` to the `sitemapPathsToIgnore` option and notice it no longer shows up in the sitemap index
7. Add `pages: [{path: '/about'}]` to the `handleSitemapRequests` config object and notice the new `/sitemap-faust-pages.xml` page show up in the sitemap index
8. Go to any existing route (e.g. `/custom-page`) to ensure it doesn't get caught by the middleware


blakewilson:
@theodesp Interesting, I hadn't seen that until now. Very similar to what were doing here, except their model has a hard constraint of 50k urls.

blakewilson:
With the current API, it is hard to include multiple indices of the same type. For example, if you had 3k posts, you would need to specify:

```tsx
{
  sitemapPathsToProxy: [
    '/wp-sitemap-posts-page-1.xml',
    '/wp-sitemap-posts-page-2.xml',
    '/wp-sitemap-posts-page-3.xml',
  ]
}
```

This gets even more complex as content is created.

Instead of `sitemapPathsToProxy`, we should probably have `sitemapPathsToIgnore`, and by default proxy any path with `sitemap` in the pathname and ends in `.xml`.

This would still allow the user to control what sitemaps get added to the index, but would require less overhead. Additionally, this would allow the user to use a wildcard in `sitemapPathsToIgnore` so the user can avoid the pitfall mentioned above. For example:

```tsx
{
  sitemapPathsToIgnore: [
    '/wp-sitemap-users-*.xml'
  ]
}
```

theodesp:
It also uses headers for cache-control. We might need to perform actual load testing here for performance.

blakewilson:
Learned that Next.js middleware's strict runtime is even stricter than I thought. It doesn't support the use of `eval` or `Function(evalString)`, which Lodash (and our packages) rely on. This is requiring us to have the `wpUrl` be manually specified, as we can't import `@faustjs/core/config` to get the proper value.

packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -213,6 +227,20 @@ export async function handleSitemapPath(
     unpairedTags: ['xml', 'xml-stylesheet'],
     processEntities: true,
     htmlEntities: true,
+    /**
```
theodesp, line 230: There is code duplication here. is there a way to share config between those two functions?

packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  if (!isUndefined(pages) && isArray(pages) && pages.length) {
+    sitemaps = [
+      ...sitemaps,
+      { loc: `${trimSlashes(origin)}/${trimSlashes(FAUST_PAGES_PATHNAME)}` },
+    ];
+  }
+
+  const wpSitemapUrl = `${trimSlashes(wpUrl)}/${trimSlashes(pathname)}`;
+  const res = await fetch(wpSitemapUrl);
```
theodesp, line 60: Will that request could be used to DDoS the Site? If for example a bot trolls the /sitemap.xml for example

packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -0,0 +1,279 @@
...
+    htmlEntities: true,
+    /**
+     * FXP can not determine if a single tag should be parsed as an array or
+     * an object, so we need to specify we always want "sitemap" tags to be an
+     * array.
+     *
+     * @see https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/2.XMLparseOptions.md#isarray
+     */
+    isArray: (tagName) => {
+      if (tagName === 'sitemap') {
```
theodesp, line 88: Simplified:
```
return tagName === 'sitemap';
```

packages/next/src/middleware/sitemaps/createSitemaps.ts
blakewilson: Yes, I suppose this could be true. This could be true of any endpoint that does a fetch request though as well. We could maybe build some internal cache that will store this requested data on a x minute interval. This could also be configured by the user.

For example: https://github.com/blakewilson/next-wp-template-hierarchy/blob/main/lib/seedQueryCache.mjs

theodesp (approved):

theodesp:
LGTM Good job!

blakewilson:
@theodesp @apmatthews Thanks yall! Changeset added.

PR #791, opened 2022-02-25
Have Next handle sitemap URL replacement

josephfusco:
## Description

These changes normalize the sitemap URLs to always have the WordPress Home URL as the base, regardless of any FaustWP setting.

Since this is essentially a reversal of the plugin filtering `post_link`, this normalization is only needed in the following sitemap providers
- posts
- taxonomies

This also fixes the issue where posts & taxonomies were not showing up in the sitemap Yoast generates due to Yoast intentionally filtering out external sitemap entries.

## Testing

1. Setup a local WordPress install with both FaustWP & Yoast (wordpress-seo) activated.
2. Visit settings -> permalinks & save changes in order to flush WordPress rewrite rules.
3. Create a few test posts.
4. Visit /sitemap.xml to view the sitemap index & go to the post sitemap.
5. You should now see all posts with the current WordPress site's URL as the base.

Repeat process for taxonomies 🔁

### Before

🟡 WordPress Core Sitemap
WordPress Core Sitemap

🟡 Yoast Sitemap
Yoast Sitemap

### After

🟢 WordPress Core Sitemap
WordPress Core Sitemap

🟢 Yoast Sitemap
Yoast Sitemap

## Documentation Changes

Should we add a new documentation section for this plugin's custom filters?

### DIFF
--- a/.changeset/wet-dots-pretend.md
+++ b/.changeset/wet-dots-pretend.md
@@ -2,4 +2,4 @@
 'faustwp': patch
 ---
 
-Normalize sitemap URLs in order to support 'partially headless' sites.
+Ensure sitemap URLs use the WordPress domain and not the headless frontend domain. Fixes a conflict with Yoast SEO that prevented post links from being added to the posts sitemap.
--- a/.changeset/wet-dots-pretend.md
+++ b/.changeset/wet-dots-pretend.md
@@ -1,0 +1,5 @@
+---
+'faustwp': patch
+---
+
+Normalize sitemap URLs in order to support 'partially headless' sites.
--- a/plugins/faustwp/includes/replacement/callbacks.php
+++ b/plugins/faustwp/includes/replacement/callbacks.php
@@ -183,13 +183,7 @@ function post_link( $link ) {
 		return $link;
 	}
 
-	$frontend_uri = faustwp_get_setting( 'frontend_uri' );
-
-	if ( $frontend_uri ) {
-		return str_replace( trailingslashit( get_home_url() ), trailingslashit( $frontend_uri ), $link );
-	}
-
-	return $link;
+	return equivalent_frontend_url( $link );
 }
 
 add_filter( 'term_link', __NAMESPACE__ . '\\term_link', 1000 );
@@ -205,16 +199,7 @@ function term_link( $term_link ) {
 		return $term_link;
 	}
 
-	$frontend_uri = faustwp_get_setting( 'frontend_uri' );
-
-	if ( empty( $frontend_uri ) ) {
-		return $term_link;
-	}
-
-	$frontend_uri = trailingslashit( $frontend_uri );
-	$site_url     = trailingslashit( site_url() );
-
-	return str_replace( $site_url, $frontend_uri, $term_link );
+	return equivalent_frontend_url( $term_link );
 }
 
 
@@ -258,5 +243,5 @@ add_filter( 'wpseo_xml_sitemap_post_url', __NAMESPACE__ . '\\yoast_sitemap_post_
  * @param string $url URL to use in the XML sitemap.
  */
 function yoast_sitemap_post_url( $url ) {
-	return normalize_sitemap_url( $url );
+	return equivalent_wp_url( $url );
 }
--- a/plugins/faustwp/includes/replacement/callbacks.php
+++ b/plugins/faustwp/includes/replacement/callbacks.php
@@ -260,13 +260,3 @@ add_filter( 'wpseo_xml_sitemap_post_url', __NAMESPACE__ . '\\yoast_sitemap_post_
 function yoast_sitemap_post_url( $url ) {
 	return normalize_sitemap_url( $url );
 }
-
-add_filter( 'wpseo_sitemap_entry', __NAMESPACE__ . '\\yoast_sitemap_entry' );
-/**
- * Filter URL entry before it gets added to the sitemap.
- *
- * @param array $sitemap_entry Array of URL parts.
- */
-function yoast_sitemap_entry( $sitemap_entry ) {
-	return normalize_sitemap_entry( $sitemap_entry );
-}
--- a/plugins/faustwp/includes/replacement/callbacks.php
+++ b/plugins/faustwp/includes/replacement/callbacks.php
@@ -228,3 +228,45 @@ function enqueue_preview_scripts() {
 	wp_enqueue_script( 'faustwp-gutenberg-filters', plugins_url( '/previewlinks.js', __FILE__ ), array(), '1.0.0', true );
 	wp_localize_script( 'faustwp-gutenberg-filters', '_faustwp_preview_link', array( '_preview_link' => get_preview_post_link() ) );
 }
+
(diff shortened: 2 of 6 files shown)


theodesp:
One thing to notice here:
* When `Enable public route redirects` is enabled then all routes will redirect to the headless site even if they show the Wordpress URL

plugins/faustwp/tests/integration/replacement/test-replacement-callbacks.php
```diff
@@ -35,8 +35,12 @@ public function test_preview_post_link_filter() {
 		$this->assertSame( 1000, has_action( 'preview_post_link', 'WPE\FaustWP\Replacement\post_preview_link' ) );
 	}
```
theodesp, line 37: Is it possible at this point to include e2e tests as well? It would be great if we had them for reference.

plugins/faustwp/includes/replacement/callbacks.php
```diff
@@ -164,34 +164,37 @@ function post_preview_link( $link, $post ) {
...
+ * @param WP_Post  $post    The post object.
  */
-function post_link( $link ) {
-	if (
-		! is_rewrites_enabled()
-		|| ( function_exists( 'is_graphql_request' ) && is_graphql_request() )
-	) {
-		return $link;
+function handle_row_actions( $actions, $post ) {
+	$frontend_uri = faustwp_get_setting( 'frontend_uri' );
```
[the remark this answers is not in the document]
theodesp: For example when I install the plugin and is_rewrites_enabled is disabled I can still see that the `view` and `preview` links point to my headless site

plugins/faustwp/tests/integration/replacement/test-replacement-functions.php
```diff
@@ -19,4 +36,24 @@ public function test_domain_replacement_enabled_returns_true_when_filtered_to_be
 		$this->assertTrue( domain_replacement_enabled() );
 		remove_filter( 'faustwp_domain_replacement_enabled', '__return_true' );
 	}
+
+	public function test_normalize_sitemap_url() {
```
apmatthews, line 40: Our other test function names tend to follow a pattern that fully spells out what exactly is being tested. In this case, it might be something like `test_normalize_sitemap_url_replaces_frontend_uri_with_home_url()`. It's long and weird, but I think it helps communicate the intent and the expectations of a specific test. It would also provide clarity if we were to add cases to test different branches of `normalize_sitemap_url()` (a test called `test_normalize_sitemap_url_does_not_replace_when_frontend_uri_not_set()`, for example). There's no hard rule on it and wouldn't consider it blocking, but just something to consider 🙂

apmatthews (commented):
This is a nice approach! Works well in my testing.

plugins/faustwp/tests/integration/replacement/test-replacement-functions.php
mindctrl: Related: if you run phpunit with the `--testdox` option, it will print a readable version of the test name. Example of the current name:
```
WPE\FaustWP\Tests\Replacement\ReplacementFunctionsTestCases
 ✔ Domain replacement enabled returns false
 ✔ Domain replacement enabled returns true when filtered to be true
 ✔ Normalize sitemap url
 ✔ Normalize sitemap entry
 ```
 
 Named like Andrew suggested (third item):
 ```
 WPE\FaustWP\Tests\Replacement\ReplacementFunctionsTestCases
 ✔ Domain replacement enabled returns false
 ✔ Domain replacement enabled returns true when filtered to be true
 ✔ Normalize sitemap url does not replace when frontend uri not set
 ✔ Normalize sitemap entry
 ```

Comment (theodesp) [reply plugins/faustwp/tests/integration/replacement/test-replacement-functions.php:40]: Dry vs Damp in unit tests is what we should follow here. I will make a note for the upcoming PR Workshop to discuss this in detail:
https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

Comment (josephfusco) [reply plugins/faustwp/tests/integration/replacement/test-replacement-functions.php]: @apmatthews I am all for having tests be more readable; Updated!
@mindctrl Thanks for the tip! I've updated our composer script to include that flag.
@theodesp Thank you!

File: plugins/faustwp/tests/integration/replacement/test-replacement-functions.php [hunk 4]
```diff
@@ -7,9 +7,26 @@
...
+
+	protected $init_settings = [
+		'frontend_uri' => 'http://localhost:3000',
+	];
+
+	public function setUp() {
+		parent::setUp();
+		update_option( $this->option, $this->init_settings );
+		// We have to clear this manually since all tests execute in a single "request"
+		$GLOBALS['wp_settings_errors'] = [];
```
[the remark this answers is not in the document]
josephfusco: Not needed. Thanks for catching that!

plugins/faustwp/includes/replacement/callbacks.php
```diff
@@ -230,3 +228,45 @@ function enqueue_preview_scripts() {
...
+ *
+ * Note that only absolute local URLs are allowed as the check after this removes external URLs.
+ *
+ * @param string $url URL to use in the XML sitemap.
+ */
+function yoast_sitemap_post_url( $url ) {
+	return normalize_sitemap_url( $url );
+}
+
+add_filter( 'wpseo_sitemap_entry', __NAMESPACE__ . '\\yoast_sitemap_entry' );
```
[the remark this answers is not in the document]
josephfusco: Great catch! Remnant of my initial work on this. Still works without it :)

plugins/faustwp/includes/replacement/functions.php
```diff
@@ -30,3 +33,41 @@ function domain_replacement_enabled() {
...
 	return apply_filters( 'faustwp_domain_replacement_enabled', is_rewrites_enabled() );
 }
+
+/**
+ * Normalizes a sitemap URL to be the original WordPress URL.
+ *
+ * @param string $url The sitemap URL to normalize.
+ * @return string
+ */
+function normalize_sitemap_url( $url ) {
```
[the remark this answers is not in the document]
josephfusco: Thanks!

I was on the fence with making this a little more general purpose as it felt out of scope so thanks for the nudge. Refactored to handle a few more replacements.

.changeset/wet-dots-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'faustwp': patch
+---
+
+Normalize sitemap URLs in order to support 'partially headless' sites.
```
apmatthews, line 5: Without any context I might wonder what exactly "normalize" means. Could we expand on that a little so it's easier for a user to asses the potential impact of this change? Maybe something like:

> Ensure sitemap URLs use the WordPress domain and not the headless frontend domain. Fixes a conflict with Yoast SEO that prevented post links from being added to the posts sitemap.

apmatthews (approved):
Nice work 🎉

.changeset/wet-dots-pretend.md
josephfusco: Wrote a commit message instead of changelog entry 😄

PR #797, opened 2022-03-01
Remove trailing slash

josephfusco:
## Description

This change sanitizes the `frontend_uri` setting from having a trailing slash `/` .

A user can attempt to save a valid URL with a trailing slash however it will be removed before the value is saved to the database. This change also accounts for those sites that already have a trailing slash in their database.

## Testing

1. Attempt to save a valid URL containing a trailing slash as the `frontend_uri`. _e.g._ `http://example.org/`
2. Save Settings
3. Verify trailing slash did not save to DB
    WP-CLI: `wp option get faustwp_settings`

Screen Shot 2022-03-01 at 5 35 48 PM

### DIFF
--- a/.changeset/cyan-apples-work.md
+++ b/.changeset/cyan-apples-work.md
@@ -1,0 +1,5 @@
+---
+'faustwp': patch
+---
+
+Remove trailing slash from frontend uri.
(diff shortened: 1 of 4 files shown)


theodesp (approved):

blakewilson (approved):
Nice solution!

blakewilson:
@josephfusco Before merging, could you please create a changeset for this?

PR #802, opened 2022-03-04
Handle `/robots.txt` requests in `handleSitemapRequests`

blakewilson:
## Description

This PR addresses handling of the `/robots.txt` route commonly needed with custom sitemap paths. This PR adds the `robotsTxt` optional property to the `handleSitemapRequests` config object.

This property is a callback function that, when defined, handles the appropriate `/robots.txt` route, and renders the text returned from this callback function:

```tsx
handleSitemapRequests(req, {
    sitemapIndexPath: '/wp-sitemap.xml',
    async robotsTxt(sitemapUrl) {
      return `
        User-agent: *
        Allow: /

        Sitemap: ${sitemapUrl}
      `;
    },
});
```

This callback also has a `sitemapUrl` param, which can be used when defining your robots.txt content.

## Testing

Tests have been added.

### Manual Testing

1. `npm run build` at the monorepo root
2. Add the following middleware to `src/pages/_middleware.ts`:
```tsx
import { handleSitemapRequests } from '@faustjs/next/middleware';
import { NextRequest, NextResponse } from 'next/server';

export default async function middleware(req: NextRequest) {
  const isSitemapRequest = await handleSitemapRequests(req, {
    sitemapIndexPath: '/wp-sitemap.xml',
    async robotsTxt(sitemapUrl) {
      return `
        User-agent: *
        Allow: /

        Sitemap: ${sitemapUrl}
      `;
    },
  });

  if (isSitemapRequest) {
    return isSitemapRequest;
  }

  return NextResponse.next();
}
```
3. Observe the `/robots.txt` route

## Screenshots

Before adding the `robotsTxt` config:

Screen Shot 2022-03-04 at 8 20 50 AM

After adding the `robotsTxt` config:

Screen Shot 2022-03-04 at 8 20 35 AM

## Dependant PRs

- #790


packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -1,3 +1,4 @@
+import { stripIndent } from 'common-tags';
```
theodesp, line 1: I this tree-shaked? I feel that you could just extract a simple function to trim the extra indentation here instead of importing a library:

https://github.com/zspecza/common-tags/blob/d3f704255f636641e80cefece3ca1c16eb773e71/src/stripIndentTransformer/stripIndentTransformer.js#L8-L31

blakewilson: @theodesp Yes! This will indeed be tree shaken. Next.js middleware operates in an [ES modules only runtime](https://nextjs.org/docs/api-reference/edge-runtime#unsupported-apis) with webpack, so only imports that are explicitly used will be bundled.

packages/next/test/middleware/sitemaps/handleSitemapRequests.test.ts
```diff
@@ -231,6 +243,61 @@ describe('handleSitemapRequests', () => {
     expect(createPagesSitemapSpy).toHaveBeenCalledWith(res, config);
   });
 
+  it('does not create /robots.txt route when robotsTxt is not defined', async () => {
+    const handleRobotsTxtSpy = jest
+      .spyOn(createSitemaps, 'handleRobotsTxt')
+      .mockImplementation();
+
+    let res = {
```
theodesp, line 251: You could mark those variables as `const`

theodesp (approved):

PR #811, opened 2022-03-15
PHP Unit test cleanup

apmatthews:
I made a couple of small tweaks to the unit tests while working on another ticket. I'm splitting them out here to keep the ticket work more focused.

- Removes an unused sample test.
- Adds aliases for test classes so IDEs stop showing errors on test methods.
- Adds a test for the post permalink filter when content replacement is enabled.
- Renames the test for the post preview link.


blakewilson (approved):
LGTM

Issue #809, opened 2022-03-14
FAUSTJS PLUGIN GIVING ERRROS AND SLOWING DOWN WEBSITE


esilverman682:
- GETTING ERROR ON ACF PRO PAGE FROM FAUSTJS PLUGIN (ERROR SEEEN BELOW)
- WORDPRESS TRIES TO LOAD LONG PERIODS OF TIME (MEMORY & RAM IS HIGH / CPU IS HIGH/ PHP 8 / PHP MEMORY HIGH -512MB)
- EVENTUALLY LOADS AFTER LONG WAIT TIME AND EVERYTHING THEREAFTER  IS FAST AND ACCESSIBLE

USING:
- FAUSTJS PLUGIN
- HEADLESS WORDPRESS
- ACF PRO
- CPTUI

ERROR: 
Warning: Undefined property: WP_Post_Type::$graphql_single_name in /var/www/html/wp-content/plugins/faustwp/includes/replacement/callbacks.php on line 151

theodesp:
Hello @esilverman682. This is a duplicate of 
https://github.com/wpengine/faustjs/issues/752

Would you mind if you provide a reproduction steps or a repo as I'm not able to verify this issue when using AFC pro?

Kind regards

theodesp:
@esilverman682 I've managed to reproduce this in PHP 8.0. Will post you updates here.

esilverman682:
I can add you to the wordpress instance & include you on the repo. What is
your github email / username? And please provide your email if you need to
look a the Wordpress instance.

Thank you.

esilverman682:
I like the faust plugin but will need to abandon it if there isn't a fix
for this error since it slows down the backend WP performance level. Have
you been able to resolve this issue

esilverman682:
Thank you for the reply. 

1. Can you provide an estimated timeframe when the new FaustWP release will be?
2. Has this error affected WP performance on your end as well?

Your response will help us on our timelines for this project and is much appreciated.

theodesp:
Hello @esilverman682. I hope you are doing well. In regards to your questions:

1. We don't have a definite answer to that. However we do have an average release cadence of 2 weeks and an engineer has already been working on providing a fix for this issue.

2. I wasn't able to verify the WP Perfomance issue and I would like to ask some questions to clarify some assumptions: 
    * Is this happening on a brand new WP install?
    * How many posts and CPTs you have on that install?
[...]

PR #812, opened 2022-03-16
Fix PHP 8 warning on non WP GraphQL post types

### LINKED ISSUE #752: Bug: Unable to create/edit Custom Fields (ACF) with FaustWP 7.3
## Applicable Versions

- WordPress version: 5.8.2 or 5.9
- FaustWP version: 7.3 (Not happening in FaustWP 7.1)

## Steps To Reproduce

1. Try to create new or edit  existing "Custom Field" through the WP dashboard

## The current behavior

Throws `Undefined property: WP_Post_Type::$graphql_single_name`
Screenshot 2022-01-28 at 09 47 29

## The expected behavior
Not erroring

apmatthews:
## Description

Post types that aren't registered with WP GraphQL (like ACF field groups) don't have a `graphql_single_name` property. As of PHP 8, accessing undefined properties on objects results in a warning rather than a notice.

This fix checks that the `graphql_single_name` property exists and is not empty before adding a `typeName` to the preview link.

## Related Issue(s):

Fixes #752
Related: #809 

## Testing

Unit test case has been added. The test case should fail in the absence of this patch regardless of what PHP version the tests are running on because our [test config](https://github.com/wpengine/faustjs/blob/9a24e06d4c932afb5a7c96fc1214064e1a7c30b0/plugins/faustwp/phpunit.xml.dist#L7-L8) converts both PHP warnings _and_ notices to exceptions. 🥳

### DIFF
--- a/plugins/faustwp/includes/replacement/callbacks.php
+++ b/plugins/faustwp/includes/replacement/callbacks.php
@@ -147,7 +147,11 @@ function post_preview_link( $link, $post ) {
 		 * the data.
 		 */
 		$post_type_object = get_post_type_object( $post->post_type );
-		if ( ! isset( $args['typeName'] ) && isset( $post_type_object ) ) {
+		if (
+			! isset( $args['typeName'] ) &&
+			isset( $post_type_object ) &&
+			! empty( $post_type_object->graphql_single_name )
+		) {
 			$gql_type_name    = ucfirst( $post_type_object->graphql_single_name );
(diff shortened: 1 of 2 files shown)


blakewilson (approved):
LGTM

esilverman682:
Does this mean that the error will still show?

blakewilson:
@esilverman682 👋

We'll be releasing v0.7.7 of the FaustWP plugin tomorrow with this fix in place.

PR #836, opened 2022-04-08
[MERL-132] Update eslint and related deps to v8

theodesp:

## Description
This PR:

 - Updates eslint in all packages to v8( core, next, react)
 - Fixes eslint react hook errors and optional chaining warnings. The changes have no real side effects.

## Testing

Perform some smoke tests in the example project. It should load pages and posts correctly.


packages/next/src/server/getProps.tsx
```diff
@@ -78,8 +78,9 @@ export async function getProps<
       async () => {
         const { cacheSnapshot: coreSnapshot } = await client.prepareReactRender(
           <RouterContext.Provider
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any,react/jsx-no-constructed-context-values
```
theodesp, line 81: There shouldn't be a need to use `useMemo` here as this call happens on the server and this hook is not available to use.

packages/next/src/components/FaustProvider.tsx
```diff
@@ -38,12 +38,14 @@ export function FaustProvider<Props = Record<string, unknown>>({
     cacheSnapshot: isNil(authSnapshot) ? undefined : authSnapshot,
   });
 
+  const value = React.useMemo(
```
blakewilson, line 41: Will introducing this have any side effects?

theodesp: No this is the recommended approach. The provider is dependant on the client which should be immutable object between each renders.

theodesp:
> @theodesp Looks good! This techinically is a code change. I wonder if we should add a changeset?

Sure I will add one.

PR #846, opened 2022-04-22
Add basic .gitattributes

josephfusco:
## Description

This adds a basic .gitattributes which allows us to specify the files and paths attributes that should be used by git when performing specific git actions.

## References

- [gitattributes Documentation](https://git-scm.com/docs/gitattributes)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,18 @@
+#
+# Set default behavior to automatically normalize line endings.
+#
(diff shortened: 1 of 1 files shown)


theodesp:
@josephfusco There is a relevant ticket in the backlog as well MERL-285 make sure you bring it into the current workload.

theodesp (approved):

PR #852, opened 2022-04-27
Explicitly set `preview=true` in preview links

apmatthews:
## Description

WordPress preview links include the `preview=true` query param by default. We rely on this query param to identify and handle preview requests on Faust.js frontends. Any WP plugin ([PublishPress](https://wordpress.org/plugins/publishpress/) is a known example) could potentially remove `preview=true`, causing previews to fail.

This PR explicitly adds `preview=true` to all preview links in case it was removed by another plugin.

## Testing

Integration test is included.

### DIFF
--- a/.changeset/light-ligers-join.md
+++ b/.changeset/light-ligers-join.md
@@ -1,0 +1,5 @@
+---
+'faustwp': patch
+---
+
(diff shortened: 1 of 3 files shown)


theodesp:
LGTM. Great!

theodesp (approved):

PR #859, opened 2022-05-02
MERL-247 multisite default settings

josephfusco:
## Description

This change handles the following multisite scenarios

- Default settings are properly applied to all sites when plugin is activated, both individual and network activations.
- Default settings are properly applied to newly created sites within a multisite network.

## Testing

1. Create a new WordPress multisite in Local. (Subdomains/subdirectories does not matter)
2. Create 2 additional sites. Call them whatever you want.
3. Going back to Local, slick "SYNC MULTISITE DOMAINS TO HOSTS FILE".
4. Install FaustWP & network activate.
5. Go to Site 1 -> Settings -> Headless to view FaustWP settings.
6. All default settings should be there.
7. Go to Site 2 -> Settings -> Headless to view FaustWP settings.
9. All default settings should be there.
10. With the plugin still active, create another site (Site 3)
11. Go to Site 3 -> Settings -> Headless to view FaustWP settings.
12. All default settings should be there.

### DIFF
--- a/plugins/faustwp/includes/utilities/callbacks.php
+++ b/plugins/faustwp/includes/utilities/callbacks.php
@@ -11,7 +11,8 @@ use function WPE\FaustWP\Detect_Conflicts\delete_conflicts_dismissed;
 use function WPE\FaustWP\Settings\{
 	get_secret_key,
 	faustwp_get_settings,
-	faustwp_update_setting
+	faustwp_update_setting,
+	maybe_set_default_settings
 };
 
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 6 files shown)


plugins/faustwp/includes/utilities/callbacks.php
```diff
@@ -31,27 +32,29 @@
...
+function handle_activation( $network_active ) {
 	if ( is_plugin_active( 'wpe-headless/wpe-headless.php' ) ) {
 		deactivate_plugins( 'wpe-headless/wpe-headless.php', true );
 	}
 
-	$secret_key = get_secret_key();
-	$settings   = faustwp_get_settings();
+	// Handle network activation of plugin within multisite.
+	if ( $network_active ) {
+		$sites = get_sites();
```
theodesp, line 46: maybe here check if the function get_sites exists:
```php
function_exists( 'get_sites' )
```

theodesp: or maybe as well place a limit:
```
$sites = get_sites( [
				'fields' => 'ids',
				'number' => PHP_INT_MAX,
			] );
```

josephfusco: @theodesp Thanks for the feedback!

I didn't include
```
function_exists( 'get_sites' )
```
as [get_sites()](https://developer.wordpress.org/reference/functions/get_sites/) was introduced WordPress 4.6 & [faustwp requires a minimum WP version of 5.7](https://github.com/wpengine/faustjs/blob/canary/plugins/faustwp/readme.txt#L4).

Good call on only selecting ids, I can add that in! In regards to explicitly setting a limit for get_sites(), WP defaults to 100; just want to clarify you're suggesting we remove that limit?

josephfusco: We can also take advantage of [wp_is_large_network()](https://developer.wordpress.org/reference/functions/wp_is_large_network/) if needed.

theodesp: Yes maybe we should consider both cases here I didn't know that.

theodesp (approved):
Awesome. Thank you @josephfusco

PR #866, opened 2022-05-11
Remove multisite subdirectory from path

josephfusco:
## Description

This fixes an issue with using wp-graphql in a WordPress multisite using subdirectories where the individual site subdirectory is included with the menu path.

## Testing

1. Create a local subdirectory multisite.
2. Add a second site in the network under `site-2`.
3. Install & network activate faustwp & wp-graphql.
4. Under site-2's headless settings, enter `http://localhost:3000` for the `frontend_uri`.
5. Run the following inside of site-2's GraphiQL.
    ```graphql
    query MyQuery {
      generalSettings {
        url
      }
      menuItems(where: {location: PRIMARY}) {
        nodes {
          label
          url
          path
        }
      }
      menus(where: {location: PRIMARY}) {
        nodes {
          menuItems {
            nodes {
              label
              url
              path
            }
          }
        }
      }
    }
    ```
6. `site-2` should no longer present within `url` or `path`. 

## Screenshots

**Before**
Screen Shot 2022-05-04 at 3 24 03 PM

**After**
Screen Shot 2022-05-04 at 3 23 49 PM

### DIFF
--- a/plugins/faustwp/includes/replacement/graphql-callbacks.php
+++ b/plugins/faustwp/includes/replacement/graphql-callbacks.php
@@ -56,14 +56,12 @@ function url_replace_recursive( &$data ) {
 		if ( ( 'url' === $key || 'href' === $key ) && is_string( $value ) ) {
 			$replacement = faustwp_get_setting( 'frontend_uri', '/' );
 			$value       = str_replace( site_url(), $replacement, $value );
-		} elseif ( is_array( $value ) ) {
-			url_replace_recursive( $value );
-		}
-
-		if ( 'path' === $key && is_multisite() ) {
+		} elseif ( ( 'path' === $key && is_multisite() ) && is_string( $value ) ) {
 			$site         = get_site();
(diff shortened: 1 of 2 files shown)


plugins/faustwp/includes/replacement/graphql-callbacks.php
```diff
@@ -59,5 +59,11 @@ function url_replace_recursive( &$data ) {
 		} elseif ( is_array( $value ) ) {
 			url_replace_recursive( $value );
 		}
+
+		if ( 'path' === $key && is_multisite() ) {
```
apmatthews, line 63: Would moving this condition up into the existing `if elseif` improve readability? I find it helpful with recursion to try and keep the recursive case as the final statement in the function.

It might also be good to keep an `is_string()` check here in case plugins extend the schema with some kind of oddball value for a field named `path`.

apmatthews (approved):
LGTM 👍🏻

PR #870, opened 2022-05-16
Update preferred node version to 16

josephfusco:
## Description

- Adds an .nvmrc file to the root of the faustjs monorepo
- Implements a matrix for testing various node versions for our packages. Currently set to 16 & 18 (latest LTS).
- Updates actions/checkout@v2 -> v3
- Updates actions/setup-node@v2 -> v3

-----
Something to note

- Node 14 comes bundled with npm 6.
- Node 15 comes bundled with npm 7. 

Since this codebase utilizes the [workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces) feature in NPM, `npm run build` in node 14 was causing an infinite loop.

If we would still like to include node 14, we will need to follow up with an alternate solution to NPM's native workspaces flag.


apmatthews (approved):
Nice!

PR #872, opened 2022-05-19
Update Docs

josephfusco:
Resolving merge conflicts from https://github.com/wpengine/faustjs/pull/502 going stale

> Changing PageIdType to URI in example since SLUG doesn't exist yet.
Fixing usePage typo


internal/website/docs/next/reference/hooks/usePost.mdx
```diff
@@ -50,7 +50,7 @@ export default function Page() {
   const { usePost } = client;
   const post = usePost({
     id: 'hello-world',
-    idType: PostIdType.SLUG,
+    idType: PostIdType.URI,
```
apmatthews, line 53: Does this part need to change? `PostIdType.SLUG` is a valid `idType` and `hello-world` _seems_ like a SLUG in the context of the other examples on the page. Neither way is technically wrong, but keeping SLUG might help avoid some confusion around what is a URI vs. what is a SLUG.

internal/website/docs/next/reference/hooks/usePage.mdx
```diff
@@ -46,9 +46,9 @@ import { client, PageIdType } from 'client';
 
 export default function Page() {
   const { usePost } = client;
-  const page = usePage({
+  const page = usePost({
```
apmatthews, line 49: Shouldn't this be `usePage` instead of `usePost`? The [original diff](https://github.com/wpengine/faustjs/pull/502/files) looked slightly different.

apmatthews (dismissed):
Looks like the old `custom-hooks.mdx` file snuck in through the merges.

blakewilson (approved):

PR #873, opened 2022-05-19
Replace unicode checkbox with SVG

apmatthews:
## Description

WordPress 6.0 contains an update to emojis that now converts a unicode checkbox we were using on the settings page to an emoji ☑️.

## Testing

This improvement isn't 6.0 specific, so you shouldn't need that installed.

1. Load the settings page with WP GraphQL installed and active and verify that the check mark renders.
2. Uninstall WP GraphQL, load the FaustWP settings page, click the "Install and Activate" button to install WP GraphQL, see that the check mark renders.

### Before on WordPress 5.9.3

Screen Shot 2022-05-19 at 3 57 55 PM

### Before on WordPress 6.0-RC3

Screen Shot 2022-05-19 at 3 54 25 PM

### After on WordPress 6.0-RC3
Screen Shot 2022-05-19 at 2 03 13 PM


josephfusco (approved):
🔥

PR #881, opened 2022-05-26
Accept user defined WP_VERSION for test containers

apmatthews:
## Description

We've been using `docker-compose up -d` to build and run a Docker application for integration and acceptance testing. This application was always built with the latest version of WordPress. Now, you can specify what version of WordPress you'd like the test application to use. The following new composer scripts are provided to help manage the test application:

```bash
composer run docker:build  # Builds and tags a WordPress docker container for the test application
composer run docker:serve  # Starts the test application and destroys the existing application if one is already running
composer run docker:start  # Build & Serve in one command
composer run docker:stop   # Destroys the existing application if one is already running
```

Most of the time you'll only need `composer run docker:start`, but the others scripts are there if you need them.

To specify the WordPress version you'd like to test against, use the WP_VERSION environment variable:

```bash
WP_VERSION=5.8 composer run docker:build
WP_VERSION=5.8 composer run docker:serve
WP_VERSION=5.8 composer run docker:start
composer run docker:stop # This one doesn't care about WP_VERSION
```

This defined the WP_VERSION for both integration AND acceptance tests. The default WP_VERSION is `latest`.

Much of this was inspired by [wp-graphql/wp-graphql](https://github.com/wp-graphql/wp-graphql). I've simplified it a bit to control scope make it easier to review.

## Testing

1. Follow the [updated `WordPress Unit Test` instructions](https://github.com/wpengine/faustjs/blob/b3e4a315fc2968b799159ce2fcf44b757d66eb42/DEVELOPMENT.md#plugins) and do NOT specify a WP_VERSION at any point.
2. Unit test output should include "Running on WordPress 6.0..." in the first few lines of output.
3. Log into the test WordPress application at `http://localhost:8080/wp-admin`
4. See that the WordPress version is 6.0
5. Follow the [updated `WordPress Unit Test` instructions](https://github.com/wpengine/faustjs/blob/b3e4a315fc2968b799159ce2fcf44b757d66eb42/DEVELOPMENT.md#plugins) and DO specify a WP_VERSION. `WP_VERSION=5.8 composer run docker:start`
6. Unit test output should include "Running on WordPress 5.8.4..."
7. Log into the test WordPress application at `http://localhost:8080/wp-admin`
8. See that the WordPress version is 5.8.4


theodesp (approved):

plugins/faustwp/composer.json
```diff
@@ -23,6 +23,16 @@
     "yoast/phpunit-polyfills": "^1.0"
   },
   "scripts": {
+    "docker:build": "bash bin/run-docker.sh build",
+    "docker:serve": [
+      "@docker:stop",
+      "docker compose up -d"
```
josephfusco, line 29: Thoughts on updating this to use the full flag since `--volumes` is being used below.
```suggestion
      "docker compose up --detach"
```

Issue #893, opened 2022-06-08
Bug: Sitemap URLs rewrite to locahost:8080 on Atlas


JEverhart383:
This issue was initially reported by a user in discord and [confirmed by me using the Faust starter using both the default WP sitemaps](https://h4xkybby1c3ydls3n5mg7bpzx.js.wpenginepowered.com/wp-sitemap.xml) and the drop in Yoast config available on the docs. When proxying sitemaps via Next.js middleware on the Atlas platform, [the value of req.url.origin is localhost:8080](https://github.com/wpengine/faustjs/blob/e85bfff965a489984b28e7aaa46afb2865bf8e06/packages/next/src/middleware/sitemaps/createSitemaps.ts#L58) and not the actual front end URL of the site, so that is what gets [rewritten when we replaceUrls](https://github.com/wpengine/faustjs/blob/e85bfff965a489984b28e7aaa46afb2865bf8e06/packages/next/src/middleware/sitemaps/createSitemaps.ts#L144).

## Applicable Versions

- `@faustjs/core` version: 0.15.6
- `@faustjs/react` version:
- `@faustjs/next` version: 0.15.6
- WordPress version: 6.0

## Steps To Reproduce

1. Create sitemap middleware based on docs
2. Deploy to Atlas platform 
3. Visit sitemap url

Link to code example:
https://github.com/JEverhart383/atlas-demo-app/blob/main/src/pages/_middleware.ts

## The current behavior

URL in sitemap entries is rewritten to localhost:8080 when proxied, b/c value of req.url.origin is localhost:8080 on Atlas platform: 
Example logs of console.log(req:NextRequest)

06/08/2022 01:30:55 PM UTC NextRequestHint [Request] {
06/08/2022 01:30:55 PM UTC   sourcePage: '/_middleware',
06/08/2022 01:30:55 PM UTC   [Symbol(internal body)]: { bodyInit: null, disturbed: false },
06/08/2022 01:30:55 PM UTC   [Symbol(internal request)]: {
06/08/2022 01:30:55 PM UTC     credentials: 'same-origin',
06/08/2022 01:30:55 PM UTC     headers: BaseHeaders [Headers] { [Symbol(map)]: [Object] },
06/08/2022 01:30:55 PM UTC     method: 'GET',
06/08/2022 01:30:55 PM UTC     referrer: 'about:client',
06/08/2022 01:30:55 PM UTC     redirect: 'follow',
[...]

PR #894, opened 2022-06-08
Replace `origin` with `protocol` + `host` in Faust sitemaps

blakewilson:
## Description

This PR replaces the [`origin`](https://developer.mozilla.org/en-US/docs/Web/API/URL/origin) URL piece with [`protocol`](https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol) + [`host`](https://developer.mozilla.org/en-US/docs/Web/API/URL/host) as `origin` may resolve to a local container (such as `localhost:8080`) on hosting providers.

## Related Issue(s):

- #893

## Testing

1. From the monorepo root, checkout the PR, and run `npm run build`
2. Add the following to `examples/next/getting-started/src/pages/_middleware.ts`:
   ```ts
    import { handleSitemapRequests } from "@faustjs/next/middleware";
    import { NextRequest, NextResponse } from "next/server";
    
    export default async function _middleware(req: NextRequest) {
      const sitemapRequest = await handleSitemapRequests(req, {
        sitemapIndexPath: `/wp-sitemap.xml`,
        pages: [
          {
            path: "/about",
          },
        ],
        async robotsTxt(sitemapUrl) {
          return `
            User-agent: *
            Allow: /
    
            Sitemap: ${sitemapUrl}
          `;
        },
      });
    
      if (sitemapRequest) {
        return sitemapRequest;
      }
    
      return NextResponse.next();
    }
   ```
3. Visit `http://localhost:3000/wp-sitemap.xml` and navigate to the index page, sitemap pages, robots.txt page, and faust pages sitemap to confirm the url replacement is happening for `http://localhost:3000`.

### DIFF
--- a/packages/next/src/middleware/sitemaps/createSitemaps.ts
+++ b/packages/next/src/middleware/sitemaps/createSitemaps.ts
@@ -54,14 +54,18 @@ export async function createRootSitemapIndex(
   req: NextRequest,
   normalizedConfig: NormalizedConfig,
 ): Promise<Response | undefined> {
+  const { pathname, host, protocol } = new URL(req.url);
+  const reqBaseUrl = `${protocol}//${host}`;
+
   const { pages, sitemapPathsToIgnore, replaceUrls, wpUrl } = normalizedConfig;
-  const { pathname, origin } = new URL(req.url);
   let sitemaps: SitemapSchemaSitemapElement[] = [];
 
   if (!isUndefined(pages) && isArray(pages) && pages.length) {
     sitemaps = [
       ...sitemaps,
-      { loc: `${trimSlashes(origin)}/${trimSlashes(FAUST_PAGES_PATHNAME)}` },
+      {
+        loc: `${trimSlashes(reqBaseUrl)}/${trimSlashes(FAUST_PAGES_PATHNAME)}`,
+      },
     ];
   }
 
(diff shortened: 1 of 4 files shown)


packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -54,14 +54,18 @@ export async function createRootSitemapIndex(
   req: NextRequest,
   normalizedConfig: NormalizedConfig,
 ): Promise<Response | undefined> {
+  const { pathname, host, protocol } = new URL(req.url);
+  const reqBaseUrl = `${protocol}//${host}`;
```
theodesp, line 58: Does this need an extra slash? (//)

packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -169,7 +173,9 @@ export function createPagesSitemap(
   req: NextRequest,
   normalizedConfig: NormalizedConfig,
 ): Response | undefined {
-  const { origin } = new URL(req.url);
+  const { host, protocol } = new URL(req.url);
+  const reqBaseUrl = `${protocol}//${host}`;
```
theodesp, line 177: Does this need an extra slash `//`?

packages/next/src/middleware/sitemaps/createSitemaps.ts
theodesp, line 58: Does this need an extra slash `//`?

packages/next/src/middleware/sitemaps/createSitemaps.ts
```diff
@@ -54,14 +54,18 @@ export async function createRootSitemapIndex(
   req: NextRequest,
   normalizedConfig: NormalizedConfig,
 ): Promise<Response | undefined> {
+  const { pathname, host, protocol } = new URL(req.url);
```
theodesp, line 57: This could be extracted as a function as we've seen it 5 times in this file.

Issue #903, opened 2022-06-15
Bug: Support for safari ios 12


marie-maxime:
It seems like faust js is not working on ios 12.

Output error: `SyntaxError: Unexpected token '?'`

NextJS supports older browsers including IE 11 and previous version of safari. I would expects faustjs to work on ios 12.

It seems like the issue is coming from a missing transpiling with optionnal chaining (`?.`)

## Applicable Versions

- `@faustjs/core` version: 0.15.6
- `@faustjs/react` version: 0.15.6
- `@faustjs/next` version: 0.15.6
- WordPress version: latest
- node: v16.14.0

## Steps To Reproduce

1. Clone starter project with faust.
```
npx create-next-app \
    -e https://github.com/wpengine/faustjs/tree/main \
    --example-path examples/next/getting-started \
    --use-npm \
    my-app
```
2. Start the project and open it in ios 12 safari ( I used browserstack). The error will be visible in the console.

## The current behavior
Faustjs crash on ios 12 & 13.

## The expected behavior
Faustjs should follows NextJS supported browser and works on ios 12.

josephfusco:
Hey @marie-maxime thanks for reporting this issue!

We have created a ticket in our internal backlog for this ([MERL-332](https://wpengine.atlassian.net/browse/MERL-332)). Will post back once we have more info!

marie-maxime:
Thank you for fixing this and addressing it as quickly! 👍

PR #913, opened 2022-06-29
Add support for iOS 12 & 13

josephfusco:
## Description

This change adds support for iOS 12 & 13 and fixes https://github.com/wpengine/faustjs/issues/903 by lowering the [target value](https://www.typescriptlang.org/tsconfig#target) in each package's TSConfig.

>The special ESNext value refers to the highest version your version of TypeScript supports. This setting should be used with caution, since it doesn’t mean the same thing between different TypeScript versions and can make upgrades less predictable.

## Testing

In order to test this on macOS, XCode should be installed with iOS 13.0 downloaded for the simulator. This likely will need to be downloaded which could take some time. https://developer.apple.com/forums/thread/110911

## Screenshots
**Before**
Screen Shot 2022-06-29 at 10 27 22 AM

**After**
Screen Shot 2022-06-29 at 10 22 02 AM

### DIFF
--- a/.changeset/sixty-fans-cough.md
+++ b/.changeset/sixty-fans-cough.md
@@ -1,0 +1,7 @@
+---
+'@faustjs/core': patch
+'@faustjs/next': patch
+'@faustjs/react': patch
+---
+
+Added support for older versions of iOS
--- a/packages/core/tsconfig-cjs.json
+++ b/packages/core/tsconfig-cjs.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "module": "commonjs",
     "outDir": "dist/cjs",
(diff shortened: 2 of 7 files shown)


blakewilson:
Nice work @josephfusco! Just tested with IOS 12 and it worked like a charm.

blakewilson (approved):

blakewilson:
Something to note for historical context: Looks like this was only an issue in dev mode. When creating an optimized production build the error goes away. Great to see finally resolved though!

marie-maxime:
@blakewilson actually, this was an issue for us on a production website hosted on vercel. Are we missing an optimization step?

PR #919, opened 2022-07-05
[FaustNX] Bug fixes to `WordPressTemplate.tsx`

theodesp:
## Description

This PR fixes the following issues that occur in `WordPressTemplate.tsx`:

1. Uses `React.cloneElement` instead of calling the component directly as we don't want to reuse the same component to render a template.
2. Handles the case when getStaticProps is used in the landing page. This is necessary to be able to view the home page at the index path `/`.


blakewilson (approved):
Nice!

PR #922, opened 2022-07-06
[MERL-167] FaustWP Enable Usage Tracking Setting

theodesp:
## Description

This PR Includes a new setting for enabling Telemetry in FaustWP. It should be disabled by default. The user should be prompted once, after install, to turn on Telemetry if they haven’t using an admin notice.

## Related Issue(s):

https://wpengine.atlassian.net/browse/MERL-167

## Testing

1. Load plugin
2. Check that the new setting is present in the Settings page and is unchecked by default.
3. Check that when you update the setting it successfully persists in the database.
4. Check that when installing the plugin there is an admin notice for prompting the user to enable Telemetry.
5. Check that when dismissing the prompt, it should not re-appear when disabling the telemetry setting.
6. Check that when deactivating and re-activating the plugin the prompt should re-appear if the telemetry setting is disabled.

## Screenshots

Screenshot 2022-07-06 at 13 16 28

Screenshot 2022-07-07 at 16 02 19


plugins/faustwp/includes/settings/callbacks.php
```diff
@@ -372,6 +374,12 @@ function display_enable_disable_fields() {
 			<input type="checkbox" id="enable_image_source" name="faustwp_settings[enable_image_source]" value="1" <?php checked( $enable_image_source ); ?> />
 			<?php esc_html_e( 'Use the WordPress domain for media URLs in post content', 'faustwp' ); ?>
 		</label>
+		<br />
+
+		<label for="enable_usage_tracking">
+			<input type="checkbox" id="enable_usage_tracking" name="faustwp_settings[enable_usage_tracking]" value="1" <?php checked( $enable_usage_tracking ); ?> />
+			<?php esc_html_e( 'Opt into anonymous usage tracking to help us make Faust.js better.', 'faustwp' ); ?>
```
josephfusco, line 381: Should we update "Faust.js" to be "FaustWP" or simply "Faust"?

josephfusco: We may want to consider dropping the period at the end of this label since the other checkbox options do not have one.

plugins/faustwp/includes/usage_tracking/callbacks.php
```diff
@@ -0,0 +1,102 @@
...
+				postData.append('action', '<?php echo esc_attr( PROMPT_ID ); ?>');
+				postData.append('nonce', '<?php echo esc_html( $nonce ); ?>');
+
+				window.fetch('<?php echo esc_url( admin_url( 'admin-ajax.php' ) ); ?>', {
+					method: 'POST',
+					body: postData,
+				})
+			});
+		}, 100);
+	});
```
josephfusco, line 78: Nice work with this @theodesp!

Although we don't have much JS in FaustWP, I'd recommend this section be moved into it's own JS file and enqueued [in the same way that FaustWP loads `wpgraphql-install.js`](https://github.com/wpengine/faustjs/blob/canary/plugins/faustwp/includes/settings/callbacks.php#L408-L433) for consistency.

In order to use PHP variables within a separate JS file, WordPress provides the function [wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/). In the case of `wpgraphql-install.js`, all localized variables have been added to a [faustwp object defined here](https://github.com/wpengine/faustjs/blob/canary/plugins/faustwp/includes/settings/callbacks.php#L431).

josephfusco (changes requested):

josephfusco (approved):
LGTM! Nice work @theodesp!

PR #926, opened 2022-07-07
FaustNX CLI POC

blakewilson:
The `faustnx` CLI will allow us to implement telemetry, and better position us in the future for other features that require our input before the `next` CLI runs.

## Description

This PR introduces a `faustnx` CLI that acts as a pass through to the `next` CLI, however, with our own telemetry that runs prior to the `next` CLI.

When running the `faustnx` CLI for the first time, you are asked if you would like to enroll in FaustNX Telemetry. By responding with either `true` or `false`, your preferences are then saved to your local machine (`~/.config/configstore/faustnx.json` on Mac) as a JSON file:

```jsonc
{
  telemetry: {
    notifiedAt: '12345', // timestamp the user was first notified about FaustNX telemetry
    anonymousId: '0000-0000-0000', // An anonymous v4 UUID for GA
    enabled: true // Either true or false based on user preferences
  }
}
```

From there, any time the `faustnx` CLI is ran, we check the users' preferences, and if telemetry is enabled, we send a GA event with relevant information.

In addition to the basic Next CLI commands like `faustnx start`, `faustnx dev`, and `faustnx build`, there is also a new command `faustnx faustnx-telemetry`, which allows the user to modify their telemetry preferences.

## Testing

1. Checkout `faustnx-cli` branch
2. Run `npm run build -w faust-nx` from the monorepo root
3. Run `npm run build -w faust-nx-cli` from the monorepo root
4. Run `npm run dev -w faust-nx-getting-started` to start the FaustNX example app in dev mode
5. Notice the FaustNX Telemetry prompt
6. Type `true`
7. Exit out of `dev` mode
8. Run `npm run build -w faust-nx-getting-started`
9. Based on your preference, you will see a telemetry event being `console.log`'d
10. Now, run `npx faustnx faustnx-telemetry` and change your preferences to `false`
11. Run `npm run build -w faust-nx-getting-started` notice that the telemetry event is not being `console.log`'d

**NOTE**: WordPress telemetry info will not be collected and will display as `undefined` since the WP telemetry endpoint doesn't exist on this branch.

## Telemetry Data

Currently the following data is being captured in a telemetry event:

```ts
export interface TelemetryData {
  wordpress: {
    faustwp_settings: {
      has_frontend_uri?: boolean;
      redirects_enabled?: boolean;
      rewrites_enabled?: boolean;
      themes_disabled?: boolean;
      image_source_replacement_enabled?: boolean;
    };
    faustwp_version?: string;
    is_wpe?: boolean;
    multisite?: boolean;
    php_version?: string;
    wp_version?: string;
  };
  node: {
    faustnx_version?: string;
    faustnx_cli_version?: string;
    apollo_client_version?: string;
    node_version?: string;
    next_version?: string;
    is_development?: boolean;
  };
}
```

### DIFF
--- a/packages/faust-nx-cli/index.ts
+++ b/packages/faust-nx-cli/index.ts
@@ -5,15 +5,20 @@ import Configstore from 'configstore';
 import { spawn } from 'child_process';
 import prompt from 'prompt';
 import { v4 as uuid } from 'uuid';
+import fs from 'fs';
 
 const GA_TRACKING_ENDPOINT = 'http://www.google-analytics.com/debug/collect';
 const GA_TRACKING_ID = 'GA-xxxx';
 const CONFIG_STORE_NAME = 'faustnx';
+const packageJson = JSON.parse(fs.readFileSync('package.json', 'utf8'));
 
 export interface TelemetryPayload {
-  cpuCount: number;
-  platform: string;
-  arch: string;
+  faustNxVersion: string;
+  faustNxCliVersion: string;
+  apolloClientVersion: string;
+  nodeVersion: string;
+  nextJsVersion: string;
+  isDevelopment: boolean;
 }
 
 /**
@@ -109,10 +114,13 @@ const sendTelemetryData = (
   }
 
   // The telemetry data to collect
-  const telemetryData = {
-    cpuCount: os.cpus().length,
-    platform: os.platform(),
-    arch: os.arch(),
+  const telemetryData: TelemetryPayload = {
+    faustNxVersion: packageJson?.dependencies?.['faust-nx'],
+    faustNxCliVersion: packageJson?.dependencies?.['faust-nx-cli'],
+    apolloClientVersion: packageJson?.dependencies?.['@apollo/client'],
+    nodeVersion: process.versions.node,
+    nextJsVersion: packageJson?.dependencies?.['next'],
+    isDevelopment: nextCliArgs[0] === 'dev',
   };
 
   /**
@@ -120,7 +128,7 @@ const sendTelemetryData = (
    * anonymousId exists, send the telemetry data.
    */
   if (
-    nextCliArgs[0] === 'build' &&
+    (nextCliArgs[0] === 'dev' || nextCliArgs[0] === 'build') &&
     config.get('telemetry.enabled') === true &&
     config.get('telemetry.anonymousId')
   ) {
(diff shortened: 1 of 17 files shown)


packages/faust-nx-cli/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (nextCliArgs[0] === 'faustnx-telemetry') {
+    const { isTelemetryEnabled } = await prompt.get({
+      properties: {
+        isTelemetryEnabled: {
+          description:
+            'Would you like to enable telemetry in FaustNX? This helps us make more informed feature decisions. true/false',
+          required: true,
+          type: 'boolean',
+        },
+      },
```
josephfusco, line 108: Noticed that lines 81-89 share the same config; thoughts on breaking this shared object out?

packages/faust-nx-cli/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+    "compilerOptions": {
+      "strict": true,
+      "outDir": "dist",
+      "module": "ES2015",
+      "target": "ES2017",
```
josephfusco, line 6: iOS folks will appreciate this 📱 🔥

packages/faust-nx/package.json
```diff
@@ -6,18 +6,20 @@
...
     "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "@apollo/client": "^3.6.6"
+    "react-dom": "^17.0.2"
   },
   "dependencies": {
+    "configstore": "^6.0.0",
     "deepmerge": "^4.2.2",
     "graphql": "^16.5.0",
     "lodash": "^4.17.21"
```
josephfusco, line 25: I might be incorrect on this in the context of SSR, but I'd consider moving `react`, `react-dom`, & `@apollo/client` to `dependencies` as they are used client side.

In addition, should our peerDependencies contain packages FaustNX already has defined as `dependencies` 
 or `devDependencies`?

josephfusco (changes requested):

packages/faust-nx-cli/index.ts
blakewilson: Good call out! Ive abstracted this into a function.

packages/faust-nx/package.json
josephfusco: Makes sense thanks @blakewilson!

theodesp (commented):
LGTM. There are some GH tests failing though.

blakewilson:
@josephfusco Thanks! @theodesp Not sure why the coverage portion of the test actions are failing. Looking into this.

PR #933, opened 2022-07-14
Add anonymous opt-in usage tracking

josephfusco:
## Description

These changes add opt-in anonymous usage tracking to Faust via CLI at build time.


plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -14,7 +14,14 @@
 	generate_refresh_token,
 	generate_access_token
 };
-use function WPE\FaustWP\Settings\get_secret_key;
+use function WPE\FaustWP\Settings\{
```
theodesp, line 17: `WPE\Compass`

plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -67,6 +75,16 @@ function rest_determine_current_user( $user_id ) {
  * @return void
  */
 function register_rest_routes() {
+	register_rest_route(
+		'faustwp/v1',
```
theodesp, line 79: `compass/v1`

PR #949, opened 2022-07-27
Merl 361 eslint config consolidation

TeresaGobble:
## Description

- Took each eslintrc.js file in the Core, Next, and React packages and added their unique options to the root level eslintrc.js file.

- Took each devDependency from the package.json files in the Core, Next, and React packages and added their eslint or prettier devDependanciesto the root level package.json file’s devDependancies

- Removed .prettierrc.js files and eslintrc.js files from Core, Next, and React packages

-  Set root-level eslintrc.js to ignore paths including ‘test’ ‘when linting (**/test/*', and     '**/tests/*’)

- Fixed a single eslint error manually in the accessToken.ts file to ensure linting script was reflecting decrease in errors correctly

- Removed “"allowJs": true,” in order to squash “Typescript error: "Cannot write file ... because it would overwrite input file” error 
[reference for this removal](https://stackoverflow.com/questions/42609768/typescript-error-cannot-write-file-because-it-would-overwrite-input-file)

## Testing

1. Checkout the MERL-361 branch
2. Run the build script 
3. Use the "lint-testing" script to see all package related linting issues (note: there will be 60 problems from Faustnx issues that are unrelated to the eslint configuration, and that will need to be examined in another ticket)
4. Please ensure no breaking changes

## Screenshots

When running "lint-testing" script, expect to see the following errors related to faustnx (that will require a separate ticket to clean!)
60 problems (59 errors,

## Documentation Changes
n/a 

## Dependant PRs
n/a


faustjs.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
```
blakewilson, line 1: We'll want to add `faustjs.code-workspace` to our `.gitignore` since this is user config.

package.json
```diff
@@ -40,10 +40,25 @@
     "changeset": "changeset",
     "version": "changeset version && node scripts/versionPlugin.js",
     "version:status": "changeset status",
-    "release": "npm run build && changeset publish"
+    "release": "npm run build && changeset publish",
+    "lint-testing": "eslint ./packages --ext js,jsx,ts,tsx --max-warnings=0 --config eslintrc.js"
```
blakewilson, line 44: We can replace `lint-testing` with the `lint` script now.

packages/core/src/gqty/index.ts
```diff
@@ -83,10 +84,10 @@ export function createAuthQueryFetcher(
       token = getAccessToken();
     }
 
-    init.headers = Object.assign(
-      { Authorization: `Bearer ${token as string}` },
-      init.headers,
-    );
+    init.headers = {
```
blakewilson, line 87: We'll want to make sure functionality is still perserved here since this is an actual code change, not just formatting.

tsconfig.json
```diff
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "allowJs": true,
```
blakewilson, line 3: I know we removed this when pair programming @TeresaGobble, but let's just double check and make sure there is no adverse effects by removing this.

blakewilson (commented):
Nice work @TeresaGobble! Excited to see this introduced. Left a few general comments.

faustjs.code-workspace
theodesp, line 1: You could remove this file as it's not needed

packages/core/src/gqty/index.ts
theodesp: Lets revert this one back as the differences are not the same:
https://stackoverflow.com/questions/32925460/object-spread-vs-object-assign

tsconfig.json
theodesp: Yeah you should restore this back.

theodesp:
@TeresaGobble great work here. You might need to follow up with the following points:
1. Resolve any PR comments
2. Resolve the PR conflicts
3. Create the follow up Ticket for fixing the linting errors.
4. Instead of creating a PR on Canary, create a PR on MERL-361 as you cannot merge in canary yet. You will have to create an empty branch based on Canary first and use Github UI to change the base.

packages/core/src/gqty/index.ts
TeresaGobble: Ah, this was one of the auto fixed items from the `--fix` command. Thank you so much for catching this guy.

theodesp:
Hey @TeresaGobble  Thank you for the update. Because the Merging is blocked and cannot do anything here unless you fix the linting errors you want to create a new branch out of this one for MERL-413 and start resolving those issues. Then we can merge MERL-413 to this branch and then merge this branch to Canary

eslintrc.js
```diff
@@ -0,0 +1,89 @@
+// core
+
+module.exports = {
```
blakewilson, line 3: This file should be deleted since we already have `.eslintrc.js`.

.eslintrc.js
```diff
@@ -0,0 +1,89 @@
+// core
```
blakewilson, line 1: Small nitpick, looks like this comment was left over from the `core` package.

blakewilson (commented):
Nice job @TeresaGobble, looks great! Left a few small comments.

blakewilson (approved):

PR #960, opened 2022-08-04
Implement experimental plugin system

blakewilson:
## Description

This PR introduces the plugin system using [@wordpress/hooks](https://www.npmjs.com/package/@wordpress/hooks) behind an `experimental` flag.

## Testing
1. Checkout the PR
2. Run `npm run build -w faust-nx`
3. Replace the `examples/next/faust-nx/faustnx.config.js` file contents with:
   ```js
    import { setConfig } from 'faust-nx';
    import templates from './wp-templates';
    
    class HelloWorld {
      apply(hooks) {
        console.log(hooks);
      }
    }
    
    /**
     * @type {import('faust-nx').FaustNXConfig}
     **/
    export default setConfig({
      templates,
      experimentalPlugins: [new HelloWorld()],
    });
   ```
4. Run `npm run dev -w faust-nx-getting-started`
5. Notice the plugin's `apply()` method gets called and the `hooks` argument gets consoled.

### DIFF
--- a/packages/faust-nx/src/config/index.ts
+++ b/packages/faust-nx/src/config/index.ts
@@ -1,6 +1,6 @@
 import { WordPressTemplate } from '../getWordPressProps.js';
 import { hooks, Plugin } from '../hooks/index.js';
-import isEqual from 'lodash/isEqual';
+import once from 'lodash/once.js';
 
 export interface FaustNXConfig {
   templates: { [key: string]: WordPressTemplate };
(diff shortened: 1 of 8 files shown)


packages/faust-nx/src/config/index.ts
```diff
@@ -1,15 +1,29 @@
...
+  // Prevent plugins from being re-initialized.
+  if (isEqual(config, _config)) {
+    return;
+  }
+
   config = _config;
+
+  const { experimentalPlugins: plugins } = _config;
+
+  plugins?.forEach((plugin) => {
```
theodesp, line 24: Will the plugins get initiated only here whenever we call the setConfig? If yes we should make sure we call this setConfig only once. preferably have something like a global variable to track it or using
https://www.geeksforgeeks.org/lodash-_-once-method/ to prevent side effects. Maybe use a warning message as well in case the user calls this twice.

josephfusco:
@blakewilson I can see the hooks argument being logged to console 👍 

What is here looks good!

josephfusco (approved):

blakewilson:
@theodesp Thoughts on something like this?

https://github.com/wpengine/faustjs/blob/merl-416-implement-hooks/packages/faust-nx/src/config/index.ts#L15-L23

theodesp (approved):

PR #961, opened 2022-08-05
[MERL 350] preview support

josephfusco:
## Description

These changes add WordPress preview support to the `faust-nx` package and it's corresponding example site.

## Testing

1. Copy `examples/next/faust-nx/.env.local.sample` to `examples/next/faust-nx/.env.local` & update to match your local WordPress site.
2. Build **faust-nx**
   ```sh
   npm run build -w faust-nx
   ```
3. Start **faust-nx-getting-started**
   ```sh
   npm run dev -w faust-nx-getting-started
   ```
4. Add some content to a page/post that has been previously published and preview
5. Notice the nextjs application resolves to /preview with preview data being displayed
6. Copy the current preview URL and paste into an incognito/private window
7. Notice the nextjs application redirect to your WordPress site to login. Once you login, you should be redirected back to the nextjs application where preview data is being shown.

## Screenshots

Screen Shot 2022-08-15 at 2 29 11 PM


packages/faust-nx/src/config/withFaust.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { previewDestination } = withFaustConfig || {};
+
+  const nextConfig = config ?? {};
+
+  const existingRedirects = nextConfig.redirects;
+  nextConfig.redirects = () =>
+    createRedirects(nextConfig, existingRedirects, previewDestination);
+
+  return nextConfig;
+}
```
theodesp, line 81: add missing new line here

packages/faust-nx/src/getWordPressProps.tsx
```diff
@@ -76,16 +80,19 @@ export async function getWordPressProps(options: getWordPressPropsConfig) {
     };
   }
 
-  if (template.query) {
+  if (template.query && !isPreview) {
     await client.query({
       query: template.query,
-      variables: template?.variables ? template.variables(seedNode) : undefined,
+      variables: template?.variables
+        ? template.variables(seedNode, false) // 2nd arg will always be false.
```
theodesp, line 87: I wonder if the second variable could be an optional `context` object with the `isPreview: false` instead of the `isPreview` directly. What do you think?

packages/faust-nx/src/hooks/usePreviewNode.ts
```diff
@@ -0,0 +1,107 @@
...
+  const node = useQuery(template?.query as DocumentNode, {
+    variables: template?.variables ? template?.variables(seedNode, true) : undefined,
+    ssr: false,
+    skip: !template?.query,
+  });
+
+  console.log({node});
+  return notFoundResponse;
+
+  // /**
```
theodesp, line 69: Are those comments needed?

examples/next/faust-nx/wp-templates/single.tsx
```diff
@@ -13,17 +13,18 @@ const Component = (props: any) => {
 
 const query = gql`
   query GetPost($uri: ID!) {
-    post(id: $uri, idType: URI) {
+    post(id: $uri, idType: URI, isPreview: $isPreview) {
```
theodesp, line 16: I wonder if the second variable could be an optional context object with the isPreview: false instead of the isPreview directly. What do you think?

packages/faust-nx/src/components/WordPressTemplate.tsx
```diff
@@ -4,33 +4,41 @@ import { getTemplate } from '../getTemplate';
...
 
-  const { __SEED_NODE__: seedNode } = props;
   const template = getTemplate(seedNode, templates);
 
   if (!template) {
     console.error('No template found');
     return null;
   }
 
+  const DEBUG_RESPONSE = usePreviewNode(template, seedNode);
```
theodesp, line 33: Is this fore debugging purposes?

packages/faust-nx/src/getWordPressProps.tsx
blakewilson: Love this idea

examples/next/faust-nx/.env.local.sample
```diff
@@ -1,2 +1,4 @@
 # GraphQL Endpoint
 NEXT_PUBLIC_GRAPHQL_ENDPOINT=http://headless.local/graphql
+NEXT_PUBLIC_WORDPRESS_URL=http://headless.local
```
theodesp, line 3: We might need to consolidate `NEXT_PUBLIC_WORDPRESS_URL` and `NEXT_PUBLIC_GRAPHQL_ENDPOINT` into one variable

examples/next/faust-nx/pages/preview.tsx
```diff
@@ -0,0 +1,12 @@
+import 'faustnx.config';
+import { getWordPressProps, WordPressTemplate } from 'faust-nx';
+import { GetServerSidePropsContext } from 'next';
+import client from 'client';
+
+export function getServerSideProps(ctx: GetServerSidePropsContext) {
+  return getWordPressProps({ client, ctx });
```
theodesp, line 7: do we need to use `getWordPressProps` as we perform everything on the client side?

packages/faust-nx/src/hooks/usePreviewQuery.ts
```diff
@@ -0,0 +1,72 @@
+// eslint-disable-next-line import/extensions
+import {useState, useEffect} from 'react'
+import { ApolloClient, InMemoryCache, ApolloQueryResult, ApolloError } from '@apollo/client';
+import type { Node } from '@faustjs/react';
+import type { DocumentNode } from 'graphql';
+import isNil from 'lodash/isNil.js';
```
theodesp, line 6: Some of those imports are not used. They should be removed
Screenshot 2022-08-17 at 13 59 40
:

packages/faust-nx/src/components/WordPressTemplate.tsx
```diff
@@ -2,40 +2,48 @@ import React, { PropsWithChildren } from 'react';
...
 export function WordPressTemplate(props: WordPressTemplateProps) {
+  const { __SEED_NODE__: seedNode, context } = props;
   const { templates } = getConfig();
 
   if (!templates) {
     throw new Error('Templates are required. Please add them to your config.');
   }
 
-  const { __SEED_NODE__: seedNode } = props;
   const template = getTemplate(seedNode, templates);
```
theodesp, line 22: will seedNode be resolved here without issues?

examples/next/getting-started/package.json
```diff
@@ -34,6 +33,7 @@
     "eslint": "^8.4.1",
     "eslint-config-next": "^12.0.7",
     "rimraf": "^3.0.2",
+    "sass": "^1.54.0",
```
blakewilson, line 36: Was this intentional?

packages/faust-nx/src/components/WordPressTemplate.tsx
```diff
@@ -2,40 +2,48 @@ import React, { PropsWithChildren } from 'react';
 import { DocumentNode, useQuery } from '@apollo/client';
 import { getTemplate } from '../getTemplate.js';
 import { SeedNode } from '../queries/seedQuery.js';
+import { WordPressTemplate } from '../getWordPressProps.js';
 import { getConfig } from '../config/index.js';
+import { usePreviewQuery } from '../hooks/usePreviewQuery.js';
 
 export type WordPressTemplateProps = PropsWithChildren<{
   __SEED_NODE__: SeedNode;
+  context?: { asPreview?: boolean };
```
blakewilson, line 11: Since this will be a prop passed down to the app's `wp-templates`, we should name it something more specific to avoid collision. Maybe somthing like `__FAUST_CONTEXT__`

packages/faust-nx/src/components/WordPressTemplate.tsx
```diff
@@ -2,40 +2,48 @@ import React, { PropsWithChildren } from 'react';
...
    * as React Hooks can not be behind conditionals
    */
-  const res = useQuery(template?.query as DocumentNode, {
-    variables: template?.variables ? template?.variables(seedNode) : undefined,
-    ssr: true,
-    skip: !template?.query,
-  });
+  let response;
+
+  if (context?.asPreview) {
```
blakewilson, line 30: We may need to take a closer look at this when the time comes to merge this into `canary` but should be fine for now.

Not sure if this violates the rules of React Hooks since `usePreviewQuery` and `useQuery` are both behind conditionals.

packages/faust-nx/src/getWordPressProps.tsx
```diff
@@ -62,6 +66,9 @@ export async function getWordPressProps(options: GetWordPressPropsConfig) {
     resolvedUrl,
   }) as DocumentNode;
 
+
+  // setup auth client
```
blakewilson, line 70: We may want to denote this as a "todo" otherwise it looks like that's what we are doing.

packages/faust-nx/src/getWordPressProps.tsx
```diff
@@ -83,17 +90,20 @@ export async function getWordPressProps(options: GetWordPressPropsConfig) {
...
+        ? template.variables(seedNode)
+        : undefined,
     });
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unsafe-return
   return addApolloState(client, {
     props: {
       __SEED_NODE__: seedNode,
+      __AS_PREVIEW__: asPreview,
```
blakewilson, line 106: This should be returned in the `context` object

packages/faust-nx/src/components/WordPressTemplate.tsx
josephfusco: I initially thought it would but React didn't throw any errors. @theodesp any thoughts here?

examples/next/faust-nx/pages/preview.tsx
josephfusco: Good catch, this was a remnant from a prior server side attempt.

packages/faust-nx/src/components/WordPressTemplate.tsx
theodesp: I get this message when I try to preview the post/page
Screenshot 2022-08-22 at 13 00 19

I think this happens because seedNode is `undefined`

examples/next/faust-nx/.env.local.sample
josephfusco: Since users have the ability to set their own graphql endpoint via WPGraphQL settings, `NEXT_PUBLIC_GRAPHQL_ENDPOINT` seems to be a hard requirement.

Grabbing the WordPress URL from `NEXT_PUBLIC_GRAPHQL_ENDPOINT` might be challenging as WordPress multisite which allows for subdirectories to be part of the WordPress URL.

blakewilson: I've added support for this in my apollo work: https://github.com/wpengine/faustjs/blob/MERL-427-plugin-apollo-client/packages/faust-nx/src/lib/getGraphqlEndpoint.ts.

@josephfusco Feel free to copy if this gets merged before my PR.

packages/faust-nx/src/components/WordPressTemplate.tsx
josephfusco: @theodesp Yes there is a separate ticket for executing the seed query on the client side ([MERL-439](https://wpengine.atlassian.net/browse/MERL-439))

packages/faust-nx/src/components/WordPressTemplate.tsx
theodesp: You will have to guarantee that only one hook gets called on each render and not the other.
You can pass on a skip parameter in `usePreviewQuery` and check within the `useEffect` if its true then return an empty response

```javascript
if (ctx?.asPreview) {
    response = usePreviewQuery(seedNode, template, false); // last boolean means do not skip
  } else {
    response = useQuery(template?.query as DocumentNode, {
      variables: template?.variables ? template?.variables(seedNode, ctx) : undefined,
      ssr: !ctx?.asPreview,
      skip: !template?.query && ctx?.asPreview,
    });
  }
```

_usePreviewQuery.ts_
```javascript
...
useEffect(() => {
 if ( skip ) {
 return previewResponse;
 }
}
...
```

PR #965, opened 2022-08-11
Setup `possibleTemplatesList` and `seedQueryDocumentNode` filters

blakewilson:
## Description

This PR introduces two filters for the plugin system:

## `possibleTemplatesList`

This filter allows a plugin to mutate the returned possible templates list to resolve a template. It can be used like:

```js
class HelloWorld {
  apply({addFilter}) {
    addFilter('possibleTemplatesList', 'faust-nx', (possibleTemplates, ctx) => {
      return [...possibleTemplates, 'new-template']
    })
  }
}
```

## `seedQueryDocumentNode`

This filter allows a plugin to mutate the returned seed query document node. We leave the implementation on how they'll mutate the seed query document node up to the user, but we should consider providing utility functions in the callback `ctx` for merging our seed query `DocumentNode` with a user provided one. It can be used like:

```js
class HelloWorld {
  apply({addFilter}) {
    addFilter('seedQueryDocumentNode', 'faust-nx', (seedQuery, ctx) => {
      return gql`
        query SeedQuery {
          ...
        }
      `
    })
  }
}
```

### DIFF
--- a/packages/faust-nx/src/getTemplate.ts
+++ b/packages/faust-nx/src/getTemplate.ts
@@ -1,8 +1,9 @@
 import { WordPressTemplate } from './getWordPressProps.js';
 import { SeedNode } from './queries/seedQuery.js';
+import { hooks } from './hooks/index.js';
 
 export function getPossibleTemplates(node: SeedNode) {
-  const possibleTemplates = [];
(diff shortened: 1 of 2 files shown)


theodesp (approved):
LGTM! Awesome!

PR #973, opened 2022-08-18
Make Apollo client configurable from the plugin system

blakewilson:
## Description

This PR abstracts the Apollo client from being specified in the app's project to being configurable from the plugin system.


examples/next/faust-nx/pages/_app.tsx
```diff
@@ -1,12 +1,11 @@
 import 'faustnx.config';
 import React from 'react';
-import client from 'client';
 import { FaustNXProvider } from 'faust-nx';
 import type { AppProps } from 'next/app';
 
 export default function MyApp({ Component, pageProps }: AppProps) {
   return (
-    <FaustNXProvider client={client} pageProps={pageProps}>
```
theodesp, line 9: I think we should still allow the user to supply their own client here irrespectively of the plugin system.

blakewilson: With the `apolloClientOptions` filter hook, they could still supply their own client that operates within the bounds of the plugin system, e.g: 

```ts
class FaustNXApolloClient {
  apply({addFilter}) {
    addFilter('apolloClientOptions', 'faust-nx', (options, ctx) => {
      return {
        ssrMode: typeof window === 'undefined',
        connectToDevTools: typeof window !== 'undefined',
        link: new HttpLink({
          uri: process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT,
          fetchOptions: {
            method: 'POST',
          },
        }),
        cache: new InMemoryCache({
          possibleTypes,
          typePolicies: {
            RootQuery: {
              queryType: true,
            },
            RootMutation: {
              mutationType: true,
            },
          },
        }),
      }
    })
  }
}
```

Do you think this is a suitable altnerative?

theodesp: Sure thats fine

theodesp (approved):
Thanks!

Issue #969, opened 2022-08-14
RSS Feeds redirect to frontend


julienkos:
I'm currently working on rewriting the feed urls from my frontend to my WP RSS Feeds to make it easier to manage and I noticed that the RSS Feed urls on the WordPress side gets redirected if the "Enable public route redirects" is checked in the Headless FaustJS Plugin settings.

I'm not sure if this is considered a bug, but from my standpoint it was probably not something that the team thought about while creating the redirects. Thanks!

Steps to reproduce:

1. Install FaustJS plugin to WordPress and fill out details in settings.
2. Check the "Enable public route redirects" checkbox.
3. Go to your WP site and add "/feed" to the url, it will redirect to your frontend app with /feed to a 404.

julienkos:
To fix this, we can just add is_feed() like this https://github.com/julienkos/faustjs/commit/a63acb75f514adc841ef1b17db9f94abea06e19c

PR #980, opened 2022-08-23
[MERL-441] WordPress RSS feeds

josephfusco:
## Description

This change prevents WordPress feeds from redirecting to the front-end application. This resolves https://github.com/wpengine/faustjs/issues/969 - props to [@julienkoss](https://github.com/julienkos) for the feedback!

## Testing

1. Pull down changes from the `MERL-441-rss` branch.
2. Go to `/feed` on your WordPress site.
3. Notice that the RSS feed for your WordPress site is accessible.


plugins/faustwp/includes/deny-public-access/callbacks.php
```diff
@@ -27,6 +27,7 @@ function deny_public_access() {
 		! is_redirects_enabled() ||
 		is_customize_preview() ||
 		doing_file_editor_save() ||
+		is_feed() ||
```
theodesp, line 30: IS there is a chance that we can add some `unit/integration` tests here?

josephfusco: Yes, we have some API testing work configured in our [Telemetry branch](https://github.com/wpengine/faustjs/pull/933/files) we can likely pull into here to help test JSON & XML responses.

theodesp (approved):
LGTM!

PR #988, opened 2022-08-29
Use node 16 in actions

blakewilson:
## Description

This PR sets the test packages action to use version `16` of Node, which it was previously using 16 and 18 (which we don't support).

Additionally, this PR addresses a small TS issue that was causing tests to fail.

### DIFF
--- a/.github/workflows/test-packages.yml
+++ b/.github/workflows/test-packages.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.version }}
+          node-version: '16'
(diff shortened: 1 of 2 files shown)


mgwright-ujet (approved):

theodesp:
@blakewilson why is Node 18 not supported BTW?

blakewilson:
@theodesp I should rephrase and say we haven't implicity added support for it yet. Our other actions run on Node 16, so this is just to follow that pattern. Additionally, Node 16 is the active supported node release, other platforms like Node 18 do not offer support yet.

We could probably offer Node 18 support, but we should ticket it and handle it appropriately.

theodesp (approved):

PR #1015, opened 2022-09-09
Add `front-page` and `home` templates

blakewilson:
## Description

This PR adds the `front-page` and `home` templates as possible templates in template hierarchy solution. This PR also drops the requirement of having a `query` in your WP Template.

### DIFF
--- a/examples/next/faust-nx/wp-templates/index.ts
+++ b/examples/next/faust-nx/wp-templates/index.ts
@@ -1,9 +1,11 @@
-import page from "./page";
-import archive from "./archive";
-import single from "./single";
+import page from './page';
+import archive from './archive';
+import single from './single';
+import frontPage from './front-page';
 
 export default {
   page,
   archive,
   single,
+  'front-page': frontPage,
 };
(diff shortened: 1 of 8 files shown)


examples/next/faust-nx/wp-templates/index.ts
```diff
@@ -1,9 +1,11 @@
...
+import page from './page';
+import archive from './archive';
+import single from './single';
+import frontPage from './front-page';
 
 export default {
   page,
   archive,
   single,
+  'front-page': frontPage,
```
josephfusco, line 10: I am wondering if we should also allow the camelCase version of the template name in addition to the traditional kebab case WordPress defines for those developers who prefer shorthand? 🤔 This would benefit those that have a large amount of custom templates. _e.g. `page-*`, `single-*`, etc._

josephfusco (approved):

josephfusco:
@blakewilson This is great! Left some non-blocking feedback :) https://github.com/wpengine/faustjs/pull/1015#discussion_r967439269

examples/next/faust-nx/wp-templates/index.ts
blakewilson: @josephfusco I really like this idea. It would avoid the awkward behavior of defining the object template key as a string. Care to create a seperate ticket for this in the backlog?

PR #1050, opened 2022-10-03
[Merl 483] Add HTTP header to indicate @faustwp usage

theodesp:
## Description

Adds `x-powered-by` header in server-side responses.

[Commit](https://github.com/wpengine/faustjs/pull/1050/commits/9f9b1c446aa079a41cb544281d9444a9489c24c3)

## Testing

Check that the server side responses contain the ` x-powered-by: FaustWP` header:

```
❯ curl http://localhost:3000/ -v
*   Trying 127.0.0.1:3000...
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< x-powered-by: Faust
< Cache-Control: no-store, must-revalidate
< ETag: "peyzputrt35e9"
< Content-Type: text/html; charset=utf-8
< Content-Length: 6996
< Vary: Accept-Encoding
< Date: Mon, 03 Oct 2022 13:00:19 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
```

**Note** This only works for server side responses and not when using `getStaticProps` as there is no way to include headers for those pages.

## Screenshots

Screenshot 2022-10-03 at 14 33 22


packages/faustwp-core/src/config/withFaust.ts
```diff
@@ -77,5 +102,8 @@ export function withFaust(
   nextConfig.redirects = () =>
     createRedirects(nextConfig, existingRedirects, previewDestination);
 
+  const existingHeaders = nextConfig.headers;
+  nextConfig.headers = () => addHeaders(nextConfig, existingHeaders);
+  nextConfig.poweredByHeader = false;
```
theodesp, line 107: Disable Next.js poweredByHeader

ChrisWiegman:
Before merging this, Check with JK. I'm not sure we'll be able to get away with "FaustWP" here as it isn't a Slug anymore.

PR #1051, opened 2022-10-03
Move and rename `axl` website

blakewilson:
## Description

This PR moves the AXL website and renames it with the legacy site to the following folder structure:
* `internal/faustjs.org`
* `internal/legacy.faustjs.org`


package.json
```diff
@@ -29,15 +29,19 @@
...
+    "docs-legacy": "npm start --prefix internal/legacy.faustjs.org",
+    "docs-legacy:build": "npm run build --prefix internal/legacy.faustjs.org",
+    "docs-legacy:clear": "npm run clear --prefix internal/legacy.faustjs.org",
+    "docs-legacy:install": "cd internal/legacy.faustjs.org && npm i && cd ..",
+    "docs-legacy:start": "npm start --prefix internal/legacy.faustjs.org",
     "test": "npm run build && npm test --workspace=@faustjs/core --workspace=@faustjs/react --workspace=@faustjs/next --workspace=@faustwp/core",
     "test:coverage": "npm run build && npm run test:coverage --workspace=@faustjs/core --workspace=@faustjs/react --workspace=@faustjs/next",
-    "start": "npm run serve:prod --prefix internal/website",
-    "wpe-build": "cd internal/website && npm i && cd .. && npm run docs:build",
+    "wpe-build": "exit 1",
```
blakewilson, line 44: Fail builds so the new docs site does not deploy on `faustjs.org`.

TeresaGobble: Ah, so it should not build docs at all (including locally) as a safety mechanism until our launch? I'm in the branch now attempting to see the new build for `npm run docs` and the other `docs` scripts in `package.lock` and am getting the following in my terminal: 

image

blakewilson: You can still build using the `docs` and `docs-legacy` scripts. However, the `wpe-build` script will return an error to not accidently deploy the site until we properly add the root directory in Atlas.

The error you are recieving seems to be because Docusaurus hasn't bee installed yet. You'll need to run `npm run docs:install` or `npm run docs-legacy:install` first.

blakewilson: However, it does look like the build is failing 😬 I'll look into that.

josephfusco (approved):
Love how the new internal directory structure reads 💯

PR #1055, opened 2022-10-04
Add preview support in example project

blakewilson:
## Description

Adds preview support in example project

Uses `databaseId` instead of `uri` in the template variables to accurately get preview data. Currently this is the bug that is causing preview data to not return when using `page`/`post` by `uri`:

https://github.com/wp-graphql/wp-graphql/issues/2091


josephfusco (approved):
Are the `package-lock.json` changes needed for this PR?
Aside from that it looks good to me 🔥

blakewilson:
@josephfusco Good call! I just reverted the changes and all tests passed

PR #1061, opened 2022-10-06
Download Changelogs

blakewilson:
## Description

This PR introduces a script in the docs site that first downloads the changelogs from the Git repo and places them in a `.changelogs` directory for consuption for our Docusaurus app. This approach has changed from pulling the changelogs from the relative directory above because we are now using an Atlas site with the root directory specified as `internal/website`. This means that our Atlas instance will no longer have access to the whole monorepo, instead just the website files.


josephfusco (approved):
LGTM!

PR #1069, opened 2022-10-11
Use `faustexample` domain for the example project wp url

blakewilson:
## Description

Update the example project `.env.local.sample` to use the https://faustexample.wpengine.com WP instance @josephfusco created.


josephfusco (approved):
LGTM!

PR #1076, opened 2022-10-12
[Merl 491] - Document `getNextServerSideProps` and `getNextStaticProps`

theodesp:
## Description

* Documents Faust Next.js file based routes using `getNextServerSideProps` and `getNextStaticProps`.
* Properly passes props to page

## Testing

* Review documentation pages at `/docs/next/reference/getNextServerSideProps` and `docs/next/reference/getNextStaticProps`.
* Navigate to `/example` path to view the file based page.

## Screenshots
Screenshot 2022-10-12 at 13 34 48

### DIFF
--- a/internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
+++ b/internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
@@ -18,8 +18,7 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
 The `getNextServerSideProps` function accepts two arguments: the server side props `context`, and an object (type `GetNextServerSidePropsConfig`) with a `Page` key.
 This should be your Next.js page component.
 
-The reason `MyPage` are passed to `getNextServerSideProps` under the hood Faust will use the Apollo Client
-to fetch the data using the 'query' and 'variables' properties. If the `Page` does not have those properties then no queries will be performed.
+Faust uses `MyPage` passed via `getNextServerSideProps` to fetch the data using the 'query' and 'variables' properties. If the `Page` does not have those properties then no queries will be performed.
 
 This is merely a convenient way to fetch the page data while reusing the logic between components. If you are interested on how this is done you can review the [code section here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/getProps.ts#L71)
 
--- a/internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
+++ b/internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
@@ -1,0 +1,41 @@
+---
+slug: /next/reference/getNextServerSideProps
+title: getNextServerSideProps
+description: The getNextServerSideProps function lets you server side render your page with WordPress data outside of the template hierarchy.
+---
+
+The `getNextServerSideProps` function lets you server side render your page with WordPress data outside of the template hierarchy using Next.js file based pages.
+The function should be returned from [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) which is required by Next.js to perform Server-Side Rendering (SSR):
+
+```tsx
+export async function getServerSideProps(context: GetServerSidePropsContext) {
+  return getNextServerSideProps(context, {
+    Page: MyPage,
+  });
+}
+```
+
+The `getNextServerSideProps` function accepts two arguments: the server side props `context`, and an object (type `GetNextServerSidePropsConfig`) with a `Page` key.
+This should be your Next.js page component.
+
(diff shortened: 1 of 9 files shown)


packages/faustwp-core/src/client.ts
```diff
@@ -104,7 +105,10 @@ export function addApolloState(
 }>['pageProps'] {
   if (pageProps?.props) {
     // eslint-disable-next-line no-param-reassign
-    pageProps.props[APOLLO_STATE_PROP_NAME] = client.cache.extract();
+    pageProps.props = {
+      ...omit(pageProps.props, ['__SEED_NODE__', '__TEMPLATE_QUERY_DATA__']),
```
theodesp, line 109: We need to properly pass on the Page `props` property so that they are available to the Page component.

blakewilson: I don't think we can omit `__SEED_NODE__` and `__TEMPLATE_QUERY_DATA__` as these are used for previews

internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
```diff
@@ -0,0 +1,41 @@
...
+    Page: MyPage,
+  });
+}
+```
+
+The `getNextServerSideProps` function accepts two arguments: the server side props `context`, and an object (type `GetNextServerSidePropsConfig`) with a `Page` key.
+This should be your Next.js page component.
+
+The reason `MyPage` are passed to `getNextServerSideProps` under the hood Faust will use the Apollo Client
+to fetch the data using the 'query' and 'variables' properties. If the `Page` does not have those properties then no queries will be performed.
```
blakewilson, line 22:
```suggestion
Faust uses `MyPage` passed via `getNextServerSideProps` to fetch the data using the 'query' and 'variables' properties. If the `Page` does not have those properties then no queries will be performed.
```

internal/faustjs.org/docs/next/reference/getNextServerSideProps.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+This is merely a convenient way to fetch the page data while reusing the logic between components. If you are interested on how this is done you can review the [code section here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/getProps.ts#L71)
```
blakewilson, line 24:
```suggestion
This is merely a convenient way to fetch the page data while reusing the logic between components. If you are interested on how this is done you can review the [source code here.](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/getProps.ts#L71)
```

internal/faustjs.org/docs/next/reference/getNextStaticProps.mdx
```diff
@@ -0,0 +1,42 @@
...
+  });
+}
+```
+
+The `getNextStaticProps` function accepts two arguments: the server side props `context`, and an object (type `GetNextStaticPropsConfig` which extends the `GetNextServerSidePropsConfig` type) with a `Page` key.
+This should be your Next.js page component or any object that contains the following properties: 'query' and `variables`.
+
+The reason `MyPage` are passed to `getNextStaticProps` is because
+under the hood Faust will use the Apollo Client to fetch the data using the 'query' and 'variables'
+properties. If the `Page` does not have those properties then no queries will be performed.
```
blakewilson, line 22:
```suggestion
Faust uses `MyPage` passed via `getNextStaticProps` to fetch the data using the 'query' and 'variables' properties. If the `Page` does not have those properties then no queries will be performed.
```

internal/faustjs.org/docs/next/reference/getNextStaticProps.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+This is merely a convenient way to fetch the page data while reusing the logic between components. If you are interested on how this is done you can review the [code section here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/getProps.ts#L35)
```
blakewilson, line 24:
```suggestion
This is merely a convenient way to fetch the page data while reusing the logic between components. If you are interested on how this is done you can review the [source code here.](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/getProps.ts#L35)
```

packages/faustwp-core/src/utils/helpers.ts
```diff
@@ -0,0 +1,7 @@
+type Obj = { [k: string]: unknown };
```
blakewilson, line 1: Do we still need this file?

blakewilson (commented):
Awesome work @theodesp! The example page is great. I left a few grammer comments

blakewilson (approved):
Nice!

PR #1096, opened 2022-10-18
Removes the `docusaurus` package since we use `@docusaurus/core` instead

blakewilson:
## Description

This PR removes the `docusaurus` dependency from the `internal/legacy.faustjs.org`/`internal/faustjs.org` packages as we use `@docusaurus/core` now.

The `package-lock.json` changes reflect the removal of the `docusaurus` package.


TeresaGobble (approved):
Pulled down and built each site, it works well! Thank you so much for this, Blake! 🙌

PR #1111, opened 2022-10-25
[MERL-527] Automate npm audit within CI/CD

josephfusco:
Signed-off-by: Joe Fusco <joe.fusco@wpengine.com>

## Description

These changes include a new workflow that will report back the results of "npm audit" using https://github.com/IBM/audit-ci/.

### DIFF
--- a/.github/workflows/audit-dependencies.yml
+++ b/.github/workflows/audit-dependencies.yml
@@ -3,7 +3,7 @@ name: Audit Dependencies
 on: pull_request
 
 jobs:
-  check_depedencies:
+  audit_depedencies:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/audit-dependencies.yml
+++ b/.github/workflows/audit-dependencies.yml
@@ -1,0 +1,12 @@
+name: Audit Dependencies
+
+on: pull_request
+
+jobs:
+  check_depedencies:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Audit for vulnerabilities
+        run: npx audit-ci@^6 --config ./audit-ci.jsonc
--- a/audit-ci.jsonc
+++ b/audit-ci.jsonc
@@ -1,0 +1,5 @@
+{
+  "$schema": "https://github.com/IBM/audit-ci/raw/main/docs/schema.json",
+  "moderate": true,
+  "allowlist": []
+}


.github/actions/compare-lockfiles/helpers.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
+
+/**
+ * Returns the parsed JSON object.
+ * @param {string} path Path to the JSON file.
+ * @returns {object}
+ */
+function parseLockfile(path) {
+  let rawdata = fs.readFileSync(`${path}/package-lock.json`);
```
theodesp, line 9: Maybe check here if the file exists here and log a message.

.github/actions/compare-lockfiles/compareLockfiles.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const core = require('@actions/core');
+const semverLt = require('semver/functions/lt');
+const { parseLockfile, getLockfilePackageVersions } = require('./helpers');
+
+const currentLockfile = parseLockfile(core.getInput('current-lockfile-path') || 'current');
```
theodesp, line 7: What happens if the `current-lockfile-path` is the same as `proposed-lockfile-path` how does this work?

.github/actions/compare-lockfiles/compareLockfiles.js
```diff
@@ -0,0 +1,38 @@
...
+  console.log('package-lock.json versions do not match');
+  process.exit(1);
+}
+
+const currentPackageVersions = getLockfilePackageVersions(currentLockfile);
+const proposedPackageVersions = getLockfilePackageVersions(proposedLockfile);
+
+const regressions = proposedPackageVersions.filter(function(proposed) {
+  const current = currentPackageVersions.find(current => current.key === proposed.key);
+  const hasRegression = semverLt(proposed.version, current.version); // proposed.version < current.version.
```
theodesp, line 21: Should we split or group `dev` vs `non-dev` dependencies and checks as we might have more that one regressions per PR? I fear that this will create a lot of false positives.

theodesp: For example a simple update create numerous changes to `node_modules/@babel` packages that change their versions using minor versions: `7.18.6` -> `7.19.6`. How could we manage those?

.github/actions/compare-lockfiles/helpers.js
theodesp, line 9: Should we check if the `package-lock.json` exists and display a message?

theodesp:
Note: The example PR is failing with a different reason:

Screenshot 2022-10-26 at 17 14 52

.github/actions/compare-lockfiles/compareLockfiles.js
josephfusco: Great catch!

.github/actions/compare-lockfiles/compareLockfiles.js
josephfusco: @theodesp The action currently looks for a lower version number following [semver](https://semver.org/) conventions. Also note that this only considers the version number of the packages within the [root packages object](https://github.com/wpengine/faustjs/blob/canary/package-lock.json#L5-L13434).

This action is not aware of any security vulnerabilities within a package, but rather just catching semver regressions.

josephfusco:
@theodesp @blakewilson Resolved those merge conflicts however the action is blocking itself from being added :D

theodesp (approved):

PR #1122, opened 2022-11-02
Telemetry notices

blakewilson:
## Description

* Adds telemetry page to legacy Faust site (faustjs.org)
* Adds privacy policy to legacy Faust site (faustjs.org)
* Fixes `dev/start/serve` scripts for Atlas


josephfusco (approved):

internal/faustjs.org/package.json
```diff
@@ -3,13 +3,12 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
+    "dev": "docusaurus start",
```
TeresaGobble, line 6: Shouldn't this script have the changelogs downloaded as well? Such as? I'm realizing this may need to be added as a second ticket, since the changelogs are only configured for the legacy site at the moment
```suggestion
    "start": "npm run downloadChangelogs && docusaurus serve",
```

TeresaGobble (approved):
Please see my one question on the changelogs. Thank you Blake!

internal/faustjs.org/package.json
blakewilson: Yes! They should, but we don't have that setup yet. [MERL-577](https://wpengine.atlassian.net/browse/MERL-577) should take care of that work.

TeresaGobble: Oh perfect! Thank you so much Blake!

PR #1124, opened 2022-11-04
CLI: Check for `faust start` command

blakewilson:
## Description

This PR fixes a bug when setting the improper `NODE_ENV` when `faust start` is ran. This PR also fixes a bug that halts CI/Node environments (when the env has no telemetry preferences set) like Atlas when running `faust start`

### DIFF
--- a/packages/faustwp-cli/index.ts
+++ b/packages/faustwp-cli/index.ts
@@ -24,6 +24,9 @@ const config = new Configstore(CONFIG_STORE_NAME);
     case 'build':
       process.env.NODE_ENV = 'production';
       break;
+    case 'start':
+      process.env.NODE_ENV = 'production';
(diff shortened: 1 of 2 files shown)


theodesp (approved):

packages/faustwp-cli/index.ts
```diff
@@ -24,6 +24,9 @@ const config = new Configstore(CONFIG_STORE_NAME);
     case 'build':
       process.env.NODE_ENV = 'production';
       break;
+    case 'start':
+      process.env.NODE_ENV = 'production';
+      break;
```
josephfusco, line 29:
```suggestion
    case 'start':
    case 'build':
      process.env.NODE_ENV = 'production';
      break;
```

PR #1126, opened 2022-11-04
Docs Site: Add "no index" feature

blakewilson:
## Description

This PR makes Docusaurus look for the `SITE_NO_INDEX` env var to determine if the site should be indexed or not on search engines. We can use this when deploying dev/unreleased docs sites. Defaults to `false` when the env var is not configured


TeresaGobble (approved):
LGTM! Thank you Blake!

matthewguywright (approved):

PR #1171, opened 2022-12-05
MERL-658: only show logs in development

matthewguywright:
## Description
Updated CLI to only log templates and seed node data in development environment

### DIFF
--- a/packages/faustwp-core/src/getTemplate.ts
+++ b/packages/faustwp-core/src/getTemplate.ts
@@ -152,10 +152,12 @@ export function getTemplate(
 
   const possibleTemplates = getPossibleTemplates(seedNode);
 
-  // Don't show console logs in PROD
-  if (process.env.NODE_ENV === 'development') {
+  if (process.env.FAUST_DEBUG_TEMPLATES) {
     // eslint-disable-next-line no-console
     console.log('possible templates: ', possibleTemplates);
+  }
+
+  if (process.env.FAUST_DEBUG_SEEDNODE) {
     // eslint-disable-next-line no-console
     console.log('seed data', seedNode);
   }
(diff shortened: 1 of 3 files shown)


packages/faustwp-core/src/getTemplate.ts
```diff
@@ -151,8 +151,14 @@ export function getTemplate(
...
-  // eslint-disable-next-line no-console
-  console.log('possible templates: ', possibleTemplates);
+
+  // Don't show console logs in PROD
+  if (process.env.NODE_ENV === 'development') {
+    // eslint-disable-next-line no-console
+    console.log('possible templates: ', possibleTemplates);
+    // eslint-disable-next-line no-console
+    console.log('seed data', seedNode);
+  }
```
josephfusco, line 161: I love the idea behind this. 

A few reservations these changes:

- Could the size of some folk's seed data require them to scroll up and past it to see possible templates?
- Is there a scenario where a user would want to see the output of possible templates and not seed node?
- A developer might want to debug a production build in a lower staging environment in which their NODE_ENV would not be `development`.

Given the above, I'm wondering if both logging possibleTemplates & seedNode can be contained to their own .env variable so that developers can toggle each one on and off as needed, regardless of NODE_ENV.

Example .env
```env
# Your WordPress site URL
NEXT_PUBLIC_WORDPRESS_URL=https://faustexample.wpengine.com

# Plugin secret found in WordPress Settings->Headless
# FAUSTWP_SECRET_KEY=YOUR_PLUGIN_SECRET

FAUST_DEBUG_TEMPLATES=1
FAUST_DEBUG_SEEDNODE=1

```

josephfusco (changes requested):
LGTM!

packages/faustwp-core/src/getTemplate.ts
blakewilson: This is a great idea @josephfusco. the seed query could be information overkill for those users that don't need it. However, creating env vars for console logs could get overwhelming as we add more down the line. What about a verbose flag in the CLI that could output more detailed information (seed query, req/res objects, etc)? 👀

matthewguywright:
Going to use verbose instead.

PR #1185, opened 2022-12-11
Improve UX for telemetry CLI

josephfusco:
## Tasks

- [X] I've filled out the WP Engine Contributor License Agreement (CLA)

## Description

The telemetry enrollment experience has been improved by eliminating the user prompt, allowing a user to explicitly enable, disable, and check the status of their enrollment in this anonymous and optional program.

These changes were heavily influenced by Next's CLI.

### ✨ New Commands ✨ 

```sh
faust telemetry status
```
```sh
faust telemetry enable
```
```sh
faust telemetry disable
```

## Testing

1. Checkout `MERL-649-telemetry-prompt-ux`
2. Install fresh set of dependencies due to changes in root package.json
3. Run `cd examples/next/faustwp-getting-started`
4. Run `../../../node_modules/.bin/faust telemetry status` to view your current status.
   screenshot of CLI running the faust telemetry status command
5. Run `../../../node_modules/.bin/faust telemetry enable` to enable telemetry.
   screenshot of CLI running the faust telemetry enable command
7. Run `../../../node_modules/.bin/faust telemetry disable` to disable telemetry.
   screenshot of CLI running the faust telemetry disable command

### DIFF
--- a/packages/faustwp-cli/index.ts
+++ b/packages/faustwp-cli/index.ts
@@ -44,7 +44,7 @@ import {
     infoLog('Faust has completely anonymous, opt-in Telemetry!');
     infoLog('You can enable it by running "npx faust telemetry enable"');
 
-    // Hydrate user's telemetry setting.
+    // Create user's telemetry setting.
     userConfig.set('telemetry', {
       notifiedAt: new Date().getTime(),
       anonymousId: uuid(),


packages/faustwp-cli/utils/handleTelemetrySubcommand.ts
```diff
@@ -0,0 +1,68 @@
...
+      log("You have opted-out of Faust.js' anonymous telemetry program.");
+      log("No data will be collected from your machine.");
+      log('Learn more: https://faustjs.org/docs/telemetry');
+      break;
+    }
+
+    case 'status': {
+      log(styles.brand('Faust.js Telemetry'));
+      printTelemetryStatus();
+      printThankYouMessage();
```
josephfusco, line 43: - [ ] Only show thank you message if telemetry is enabled

packages/faustwp-cli/index.ts
```diff
@@ -38,30 +35,11 @@ const config = new Configstore(CONFIG_STORE_NAME);
...
-    !config.all?.telemetry ||
-    config.all?.telemetry?.enabled === undefined ||
-    !config.all?.telemetry?.anonymousId
-  ) {
-    /**
-     * Do not prompt for telemetry if preferences are not set and the command
-     * that is being ran is build or start. We do not want to halt the build of a
-     * production site that likely does not have preferences saved.
-     */
-    if (arg1 !== 'build' && arg1 !== 'start' && !disableCliInteraction()) {
```
blakewilson, line 56: If prefs are not saved, we may want to `console.log` a statement telling them that telemetry exists. I had a WIP PR doing something like this. Maybe we can incorporate that message here?

https://github.com/wpengine/faustjs/pull/1176

packages/faustwp-cli/utils/handleTelemetrySubcommand.ts
```diff
@@ -0,0 +1,74 @@
+import chalk from 'chalk';
+import { getCliArgs } from './getCliArgs.js';
+import { userConfig } from './userConfig.js';
+import { styles } from './styles.js';
+
+const TELEMETRY_ENABLED = 'telemetry.enabled';
+const { log } = console;
```
blakewilson, line 7: This feels like it may add confusion down the road. Any reason why we would not use just `console.log()`?

packages/faustwp-cli/utils/handleTelemetrySubcommand.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    case 'status': {
+      log(styles.brand('Faust.js Telemetry'));
+      logTelemetryStatus();
+      log('Faust.js telemetry is completely anonymous.');
+      logTelemetryDocsUrl();
+      break;
+    }
+
+    default: {
```
blakewilson, line 56: It looks like running:

```
npx faust telemetry
```

Results in `"undefined" is not a valid telemetry subcommand`. If there is no subcommand, we may want to just show the status.

packages/faustwp-cli/utils/promptUserForTelemetryPref.ts
```diff
@@ -1,37 +0,0 @@
...
-
-  if (isInit) {
-    config.set('telemetry', {
-      notifiedAt: new Date().getTime(),
-      anonymousId: uuid(),
-      enabled: isTelemetryEnabled,
-    });
-  } else {
-    config.set('telemetry.enabled', isTelemetryEnabled);
-  }
```
blakewilson, line 36: This functionality was removed but is important.

We are no longer determining if it's the user first time setting preferences. We need this to happen so we can assign them a unique id, set the notified time, etc.

Since we are no longer prompting, and we don't have a clear delineation if we have notified the user yet, maybe we do the following:
1. Check if the user has preferences
2. If not, show them a message saying Faust has telemetry, and create their preferences with `enabled: false`.
3. Now that we have their preferences and when they were notified, we can stop showing them the above mentioned console log

packages/faustwp-cli/utils/validateFaustEnvVars.ts
```diff
@@ -12,13 +11,8 @@ export const validateFaustEnvVars = () => {
...
-    warnLog('Could not find FAUSTWP_SECRET_KEY environment variable.');
-    warnLog('Some functionality may be limited.');
-  }
+    errorLog('Could not find FAUSTWP_SECRET_KEY environment variable.');
 
-  if (disableCliInteraction()) {
-    infoLog(
-      'FAUST_NO_INTERACTION is set. Specify `false`, `0`, or remove to disable.',
-    );
+    process.exit(0);
```
blakewilson, line 16: We've never explicitly told people that the `FAUSTWP_SECRET_KEY` was a requirement. This would be a pretty big breaking change.

I think we should consider this a bit more before introducing.

internal/faustjs.org/docs/telemetry.mdx
```diff
@@ -4,18 +4,17 @@ title: Telemetry
...
+Faust collects completely anonymous telemetry data about general usage. Participation in this anonymous program is optional.
 
-## What is being collected?
+## What data do we collect?
 
-Once you opt-in, faust will collect the following information at build time:
-
-- Which settings are enabled/disabled in the Faust plugin
-- Faust plugin version
-- If the WordPress site is hosted on WP Engine
```
blakewilson, line 15: This was removed but we do collect this.

packages/faustwp-cli/index.ts
```diff
@@ -38,30 +35,11 @@ const config = new Configstore(CONFIG_STORE_NAME);
...
-      await promptUserForTelemetryPref(true, config);
-    }
-  }
 
   // eslint-disable-next-line default-case
   switch (arg1) {
-    case 'faust-telemetry': {
-      await promptUserForTelemetryPref(false, config);
```
blakewilson, line 64: We are no longer initializing user preferences, see my comment below.

packages/faustwp-cli/utils/shouldFireTelemetryEvent.ts
```diff
@@ -0,0 +1,20 @@
+import { getCliArgs } from './getCliArgs.js';
+import { userConfig } from './userConfig.js';
+
+export function shouldFireTelemetryEvent() {
```
blakewilson, line 4: Nice 💯

blakewilson (changes requested):
Nice @josephfusco! I'm excited for the interactive prompt to be replaced by the new CLI options. Also, love the rename from `faust-telemetry` to just `telemetry`. Left a few requests to avoid breaking changes. Also, looks like there is some lingering linting errors.

.changeset/dull-dots-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/cli': minor
+---
+
+Remove user prompt from telemetry command.
```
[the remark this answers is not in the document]
josephfusco: 💯

packages/faustwp-cli/index.ts
josephfusco: Added! Nice work with #1176!

packages/faustwp-cli/utils/handleTelemetrySubcommand.ts
josephfusco: Great call! Updated to display the telemetry status when no subcommand exists

package.json
```diff
@@ -46,8 +46,8 @@
     "version": "changeset version && node scripts/versionPlugin.js",
     "version:status": "changeset status",
     "release": "npm run build && changeset publish",
-    "lint": "eslint ./packages --ext js,jsx,ts,tsx --max-warnings=0 --config .eslintrc.js",
-    "lint:fix": "eslint ./packages --ext js,jsx,ts,tsx --max-warnings=0 --config .eslintrc.js --fix"
```
josephfusco, line 50: This change enables individual packages to override the base eslint rules.

In the case of this pull request, it is appropriate to allow the usage console.log() `@faustwp/cli` (node environment), but not within `@faustwp/core` (browser)

packages/faustwp-cli/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
```
blakewilson, line 1: Just to be sure, specifying a nested ESLint config will still use the parent rules in the monorepo root as long as they are not overwritten right?

packages/faustwp-cli/utils/handleTelemetrySubcommand.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export const handleTelemetrySubcommand = () => {
+  const telemetryEnabled = userConfig.get(TELEMETRY_ENABLED);
+  const args = getCliArgs();
+  const subcommand = args[1];
+
+  switch (subcommand) {
+    case 'enable': {
+      userConfig.set(TELEMETRY_ENABLED, true);
+      console.log(styles.success('Done!'));
```
blakewilson, line 44: We should call `logUserPrefsHaveBeenSaved()` for when telemetry is enabled also.

blakewilson (commented):
Nice work @josephfusco! Just a few final questions/suggestions.

packages/faustwp-cli/.eslintrc
josephfusco: Correct, this package should inherit all parent rules defined higher up in the project.

For future reference, if we did want a unique ESLint config for the lets say the CLI, adding
```
{
  root: true
}
```

to a nested ESLint will inform ESLint to NOT inherit those rules.

https://eslint.org/docs/latest/user-guide/configuring/configuration-files#cascading-and-hierarchy

blakewilson: Awesome! Good to know for the future

blakewilson (approved):
Awesome work @josephfusco!

PR #1220, opened 2023-01-17
Add verbose mode to Faust

blakewilson:
## Tasks

- [ x] I have signed a [Contributor License Agreement (CLA)](https://github.com/wpengine/faustjs#contributor-license-agreement) with WP Engine.

## Description

Adds a verbose mode to Faust by specifying the `FAUST_VERBOSE` environment variable.

* Removed the `disableLogging` option from the @faustwp/core logger. This was a reminant of old Faust and hasn't been explicitly documented in new Faust so it's not a breaking change.
* Faust now only sets the `NODE_ENV` if it wasn't already set. That way if users wanted custom `NODE_ENV`s we are not overriding their preferences.

Currently there is a logger in both @faustwp/core and @faustwp/cli, making the code not as DRY as it could be. However, since we may be researching how we can built the CLI into core I figure we could address this at a later time. I didn't want to create exports in the CLI just for logger utils for core to use.

## Screenshots

Screenshot 2023-01-17 at 3 28 03 PM

Much cleaner...

### DIFF
--- a/packages/faustwp-cli/index.ts
+++ b/packages/faustwp-cli/index.ts
@@ -15,28 +15,38 @@ import {
   validateFaustEnvVars,
   userConfig,
   infoLog,
+  verboseLog,
+  isVerbose,
 } from './utils/index.js';
 
 // eslint-disable-next-line func-names, @typescript-eslint/no-floating-promises
 (async function () {
   const arg1 = getCliArgs()[0];
 
-  switch (arg1) {
-    case 'build':
-      process.env.NODE_ENV = 'production';
-      break;
-    case 'start':
-      process.env.NODE_ENV = 'production';
-      break;
-    case 'test':
-      process.env.NODE_ENV = 'test';
-      break;
-    case 'dev':
-    default:
-      process.env.NODE_ENV = 'development';
-      break;
-  }
   dotenv.config();
+
+  if (isVerbose()) {
+    verboseLog('Faust is running in verbose mode');
+  }
+
+  if (!process.env.NODE_ENV) {
+    switch (arg1) {
+      case 'build':
+        process.env.NODE_ENV = 'production';
+        break;
+      case 'start':
+        process.env.NODE_ENV = 'production';
+        break;
+      case 'test':
+        process.env.NODE_ENV = 'test';
+        break;
+      case 'dev':
+      default:
+        process.env.NODE_ENV = 'development';
+        break;
+    }
+  }
+
   validateFaustEnvVars();
 
   // Inform user of telemetry program.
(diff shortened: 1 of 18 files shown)


packages/faustwp-cli/index.ts
```diff
@@ -15,28 +15,38 @@ import {
...
-      process.env.NODE_ENV = 'development';
-      break;
-  }
   dotenv.config();
+
+  if (isVerbose()) {
+    verboseLog('Faust is running in verbose mode');
+  }
+
+  if (!process.env.NODE_ENV) {
```
blakewilson, line 32: Checks first to see if `NODE_ENV` has been set. If it has, do not modify it.

packages/faustwp-cli/utils/isVerbose.ts
```diff
@@ -0,0 +1,9 @@
+export function isVerbose(): boolean {
+  const isVerboseEnvVar = process.env.FAUST_VERBOSE;
```
blakewilson, line 2: Should we call it `FAUST_VERBOSE` or `FAUST_DEBUG`?

packages/faustwp-core/package.json
```diff
@@ -20,15 +20,16 @@
...
     "rimraf": "^3.0.2",
     "ts-jest": "^27.0.7",
     "ts-loader": "^9.2.6",
-    "typescript": "^4.4.4",
-    "fetch-mock": "9.11.0"
+    "typescript": "^4.4.4"
   },
   "dependencies": {
     "@wordpress/hooks": "^3.14.0",
+    "chalk": "^4.1.2",
```
blakewilson, line 32: Using Chalk >5 because it is ESM only.

blakewilson: Also, this dependency won't effect bundle size much since @faustwp/core requires @faustwp/cli, and @faustwp/cli has a dependency of chalk, so it will be deduped.

packages/faustwp-core/src/getWordPressProps.tsx
```diff
@@ -69,12 +70,19 @@ export async function getWordPressProps(options: GetWordPressPropsConfig) {
...
 
+  verboseLog(`Seed Node for resolved url: "${resolvedUrl}": `, seedNode);
+
   if (!seedNode) {
     return {
       notFound: true,
     };
   }
 
+  infoLog(
```
blakewilson, line 81: Explicitly calling this here so it only runs once on the server, instead of on the client too.

packages/faustwp-cli/utils/isVerbose.ts
josephfusco: +1 for `FAUST_DEBUG` to match WordPress' usage of `WP_DEBUG`

blakewilson: I tend to agree. Will make the change!

blakewilson: Updated 👍 Everything now references "debug" opposed to "verbose" now.

josephfusco (approved):
Tested this locally and it works great @blakewilson 🔥

PR #1259, opened 2023-02-08
Merl 365 faust toolbar

josephfusco:
## Tasks


packages/faustwp-core/package.json
```diff
@@ -5,6 +5,18 @@
...
+  "exports": {
+    ".": {
+      "require": "./dist/cjs/index.js",
+      "import": "./dist/mjs/index.js"
+    },
+    "./toolbar": {
+      "require": "./dist/cjs/components/Toolbar/index.js",
+      "import": "./dist/mjs/components/Toolbar/index.js"
+    },
+    "./dist/css/toolbar.css": "./dist/css/toolbar.css",
```
blakewilson, line 17: Instead of exporting the style, we should include it outselves in the toolbar component

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+  useEffect(() => {
+    const filteredNodes = hooks.applyFilters('toolbarNodes', coreToolbarNodes, {
+      client,
+      seedNode,
+    }) as FaustToolbarNodes;
+
+    const uniqueKeys = new Set(filteredNodes.map((nodes) => nodes.key));
+
+    if (uniqueKeys.size < filteredNodes.length) {
+      throw new Error('Toolbar Nodes must have unique keys.');
```
blakewilson, line 95: Nice! We should also surface the `uniqueKeys` so users can easily determine what keys are not unique

packages/faustwp-core/src/components/Toolbar/ToolbarSubmenu.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { PropsWithChildren } from 'react';
+
+interface Props extends PropsWithChildren<any> {
+  id?: string;
+  children: React.ReactNode | undefined;
+  secondary?: boolean;
+}
```
blakewilson, line 7:
```suggestion
type Props = PropsWithChildren<{
  id?: string;
  secondary?: boolean;
}>
```

packages/faustwp-core/src/utils/adminUrl.ts
```diff
@@ -0,0 +1,13 @@
+import { untrailingslashit } from './untrailingslashit.js';
+
+/**
+ * Retrieves the URL to the admin area for the current site.
+ *
+ * @link https://developer.wordpress.org/reference/functions/admin_url/
+ */
+export function adminUrl(path = ''): string {
+  const baseUrl: string = process.env.NEXT_PUBLIC_WORDPRESS_URL || '';
+  const sanitizedBaseUrl = untrailingslashit(baseUrl);
```
blakewilson, line 10: You can use Lodash's `trim` to accomplish this:
```
import trim from 'lodash/trim.js'
```

```suggestion
  const sanitizedBaseUrl = trim(baseUrl, '/');
```

packages/faustwp-core/src/utils/adminUrl.ts
```diff
@@ -0,0 +1,13 @@
+import { untrailingslashit } from './untrailingslashit.js';
+
+/**
+ * Retrieves the URL to the admin area for the current site.
+ *
+ * @link https://developer.wordpress.org/reference/functions/admin_url/
+ */
+export function adminUrl(path = ''): string {
+  const baseUrl: string = process.env.NEXT_PUBLIC_WORDPRESS_URL || '';
```
blakewilson, line 9: We should also get the `baseUrl` from `getWpUrl()` to respect the filters we have in place.

packages/faustwp-core/src/utils/untrailingslashit.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
blakewilson, line 1: We can remove this and use `import trim from 'lodash/trim.js'` instead

packages/faustwp-core/src/wpHooks/overloads.ts
```diff
@@ -68,6 +69,16 @@ type FaustCoreFilters = {
     callback: (wpUrl: string, context: Record<string, never>) => string,
     priority?: number | undefined,
   ): void;
+
+  addFilter(
+    hookName: 'toolbarNodes',
+    namespace: string,
+    callback: (
+      toolbarNodes: FaustToolbarNodes,
+      context: Record<string, never>,
```
blakewilson, line 78:
```suggestion
      context: { seedNode: SeedNode },
```

packages/faustwp-core/src/config/index.ts
```diff
@@ -8,6 +8,7 @@ import { hooks, FaustPlugin } from '../wpHooks/index.js';
 
 export interface FaustConfig {
   templates: { [key: string]: WordPressTemplate };
+  experimentalToolbar: boolean;
```
blakewilson, line 11:
```suggestion
  experimentalToolbar?: boolean;
```

blakewilson (changes requested):
Nice work on this @josephfusco! Can't wait to see this in prod

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
josephfusco: @blakewilson Appreciate the feedback! 

This guard clause prevents the app from running if a user attempts to add a Toolbar node with an existing key (via @wordpress/hooks). Surfacing `uniqueKeys` feels unnecessary here but I could be misunderstanding a scenario.

packages/faustwp-core/src/utils/adminUrl.ts
josephfusco: Absolutely! Good catch!

packages/faustwp-core/src/components/FaustProvider.tsx
```diff
@@ -1,15 +1,37 @@
...
   const apolloClient = useApollo(pageProps);
 
-  return <ApolloProvider client={apolloClient}>{children}</ApolloProvider>;
+  return (
+    <ApolloProvider client={apolloClient}>
+      {experimentalToolbar && (
+        <Toolbar
+          key={`faust-toolbar-${router.asPath}`} // Required in order to load each route's own seed node.
+          // eslint-disable-next-line no-underscore-dangle
+          seedNode={pageProps.__SEED_NODE__ as SeedNode}
```
blakewilson, line 31: We don't want to cast this property as `SeedNode` since it can be `undefined`. Instead, we should denote in the `Toolbar` props that `__SEED_NODE__` may be `undefined`

```suggestion
          seedNode={pageProps.__SEED_NODE__}
```

packages/faustwp-core/src/wpHooks/overloads.ts
```diff
@@ -68,6 +69,16 @@ type FaustCoreFilters = {
     callback: (wpUrl: string, context: Record<string, never>) => string,
     priority?: number | undefined,
   ): void;
+
+  addFilter(
+    hookName: 'toolbarNodes',
+    namespace: string,
+    callback: (
+      toolbarNodes: FaustToolbarNodes,
+      context: { seedNode: SeedNode },
```
blakewilson, line 78: Since `seedNode` could be undefined:

```suggestion
      context: { seedNode?: SeedNode },
```

packages/faustwp-core/src/lib/getWpUrl.ts
```diff
@@ -1,10 +1,14 @@
 import trim from 'lodash/trim.js';
 import { hooks } from '../wpHooks/index.js';
 
-export function getWpUrl(): string {
+export function getWpUrl(path = ''): string {
```
blakewilson, line 4: 💯

packages/faustwp-core/src/components/Toolbar/nodes/MyAccount.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  if (!isReady) {
+    return <>Loading...</>;
+  }
+
+  if (isAuthenticated === true) {
+    return <AuthenticatedAccount />;
+  }
+
+  return <a href={loginUrl as string}>Login</a>;
```
blakewilson, line 107: We likely want to `return null` here since the user needs to be authenticated to see the Faust toolbar

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
blakewilson: I was just thinking in the case, for example, a user downloads a toolbar node plugin from npm and installs it, they may not necessarily know what unique IDs are colliding since they didn't built the toolbar plugin themselves. Knowing which IDs collided will allow them to notify the package maintainer of the issue.

packages/faustwp-core/src/components/Toolbar/nodes/Edit.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export function Edit({ seedNode }: Props) {
+  if (
+    seedNode === undefined ||
+    seedNode?.isFrontPage ||
+    seedNode?.isPostsPage
+  ) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return <></>;
```
blakewilson, line 17:
```suggestion
    return null;
```

packages/faustwp-core/src/components/Toolbar/ToolbarSubmenuWrapper.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type Props = {
+  children: React.ReactNode | string;
+};
```
blakewilson, line 5:
```suggestion
import React, { PropsWithChildren } from 'react';

type Props = PropsWithChildren<{}>;
```

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Handle adding `admin-bar` body class on render.
+   *
+   * This could eventually be handled with a Faust bodyClass hook
+   * @link https://developer.wordpress.org/reference/hooks/body_class/
+   */
+  useEffect(() => {
+    if (!isAuthenticated) {
+      return undefined;
```
blakewilson, line 128:
```suggestion
      return;
```

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,181 @@
...
+ */
+export type FaustToolbarContext = {
+  seedNode: SeedNode;
+};
+
+/**
+ * Toolbar props.
+ */
+export type ToolbarProps = {
+  seedNode: SeedNode;
```
blakewilson, line 60:
```suggestion
  seedNode?: SeedNode;
```

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * An array of toolbar nodes.
+ */
+export type FaustToolbarNodes = FaustToolbarNode[];
+
+/**
+ * Toolbar context.
+ */
+export type FaustToolbarContext = {
+  seedNode: SeedNode;
```
blakewilson, line 53:
```suggestion
  seedNode?: SeedNode;
```

blakewilson:
Nice work @josephfusco! A few more things have surfaced around the seed node being optional. I think there is that, and then figuring how we can package styles automatically, and then this is ready to go!

packages/faustwp-core/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,183 @@
...
+        component: <MyAccount />,
+        additionalClassNames: 'with-avatar',
+      },
+    ];
+  }, [seedNode]);
+
+  const [toolbarNodes, setToolbarNodes] = useState(coreToolbarNodes);
+  let { isAuthenticated } = useAuth();
+
+  isAuthenticated = true;
```
blakewilson, line 99: Possible remnant of testing that should be removed

josephfusco: Yes, thank you!

blakewilson (approved):
Well done @josephfusco! Can't wait to get this into the hands of users

PR #1278, opened 2023-02-20
Merl-751: require authentication to show the toolbar

matthewguywright:
## Description
Required authentication to see and use the toolbar.

## Testing
Ensure if authenticated that you can see the toolbar if it's set to true in the config. If you are not authenticated then you should not see it even if it is set to true in the faust.config.js.


theodesp:
Hey @matthewguywright  this should merge into the `MERL-365-faust-toolbar` branch instead of `canary`

PR #1284, opened 2023-02-22
MERL-751

matthewguywright:
## Description
Require authorized user to view toolbar in headless site.

## Testing
If experimentalToolbar=true, and logged in, you should see the toolbar. Otherwise you should never see it.


packages/faustwp-core/src/components/FaustProvider.tsx
```diff
@@ -21,10 +22,11 @@ export function FaustProvider(props: {
   const { experimentalToolbar } = getConfig();
   const router = useRouter();
   const apolloClient = useApollo(pageProps);
+  const { isAuthenticated } = useAuth();
 
   return (
     <ApolloProvider client={apolloClient}>
-      {experimentalToolbar && (
+      {experimentalToolbar && isAuthenticated && (
```
blakewilson, line 29: Let's add `isAuthenticated` to the `Toolbar` instead of the `FaustProvider`. Since there are side effects of calling `useAuth` (it makes a request for the token endpoint), this request will always be made regardless of the value of `experimentalToolbar`. Moving `useAuth` into `Toolbar` will resolve this.

blakewilson (changes requested):

blakewilson (approved):

PR #1289, opened 2023-02-23
Style: Use `div` for `DefaultBlock` instead of `span`

theodesp:
## Tasks

## Description

Changes the default element for `DefaultBlock` from `<span>` to `<div>`.

## Testing

* Use `WordPressBlocksViewer` to render a list of blocks that are not implemented. The final markup should contain divs for fallback blocks instead of spans.


josephfusco (approved):
🔥🔥

Issue #1287, opened 2023-02-22
Bug: Endless refreshing when trying to preview post with different basePath in Next.js config


riderjensen:
When adding a `basePath` to `next.config.js` (for example `/blog`), the preview mode gets stuck in an endless authentication loop. 

## Applicable Versions

- `@faustjs/core` version: 0.2.4
- WordPress version: 6.1.1

## Steps To Reproduce

1. Create a base wordpress instance with faust/gql/etc
2. Create a faust project with base config for next.js
3. In the `next.config.js` add a property called `basePath` and set it to `/blog`
4. In wordpress, in your Faust settings, set `Front-end site URL` to reflect the basePath (i.e. `http://localhost:3000/blog`)
5. Attempt to preview one of the posts
6. Note the page never loads and refreshes over and over again with a different code query parameter

Link to code example:

While I don't have a code sample, I can point out the reason this is happening:

1. Observe that the `FAUST_API_BASE_PATH` is hardcoded within the libs [here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/lib/constants.ts#L1)
2. Observe that the `FAUST_API_BASE_PATH` is imported and used within the router [here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/server/router/index.ts#L34)
3. Observe that the `FAUST_API_BASE_PATH` is imported and used within the accessToken.ts file [here](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/auth/client/accessToken.ts#L134)
4. Note that there is no way to change this constant in any config 
5. It should be noted that the [docs](https://faustjs.org/docs/next/reference/api-router) for trying to change the `apiBasePath` do not apply here as you can see the [config](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/config/index.ts#L9) takes no parameter for `apiBasePath`
7. [accessToken.ts](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/auth/client/accessToken.ts#L133) is called by [authorize.ts](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/auth/authorize.ts#L47) which is called by [WordPressTemplate](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/components/WordPressTemplate.tsx#L71) on mount.
[...]

PR #1290, opened 2023-02-23
fix[faustwp-core]: (1287) fix endless refreshing on preview and add in two faust configuration variables

riderjensen:
## Tasks

## Description

This MR does a few things
- Allow users to configure a `basePath` in `faust.config.js`.
  - Users can now have a front-end url like `https://example.com/blog` and still have preview mode work
- Allow users to configure a `baseApiPath` in `faust.config.js`.
  - Users can now change their api folder structure from the hardcoded `/api/faust`.
  - Now we set a default for `baseApiPath` within the configuration and allow users to overwrite it.
  - This functionality was present in `core` but not yet migrated over.
- Documentation was added for `basePath` and updated for `baseApiPath` to correctly reflect the code
- Removes `Partial` from the config object return type in Faust
  - The reasoning behind this is that we are setting default values for a few items which means returning Partial actually lies about some values not being present.

## Related Issue(s):

- #1287

## Testing

See more detailed information on the bug ticket but tests were done locally by configuring a basePath in `next.config.js` and changing the front-end URL in the faust config.

## Screenshots

No visual changes were included

## Documentation Changes

See updated documentation 
- internal/faustjs.org/docs/faustwp/settings.mdx
- internal/faustjs.org/docs/next/reference/api-router.mdx

## Dependant PRs

None

## Thoughts

As part of the setup for Faust.js and [connecting your wordpress site](https://faustjs.org/docs/getting-started#connecting-your-wordpress-site), `NEXT_PUBLIC_WORDPRESS_URL` is added to the `.env` file. I am wondering if it might make more sense to have that be our basis for `basePath`. 

For example, on startup, we can pull that URL from the .env and parse it using `packages/faustwp-core/src/utils/convert.ts` and then set the basePath ourselves. This is less work for the user as we can just tell them to match the `NEXT_PUBLIC_WORDPRESS_URL` in the `.env` and the `Front-end site URL` from the Faust plugin.

We could even remove basePath and simply have a non-editable field of `baseUri` which we could then use for more specificity in the fetch functions for example in [accessToken.ts](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/auth/client/accessToken.ts#L142).


blakewilson:
Hey @riderjensen, thanks for this PR! We really appreciate these community contributions to make Faust better.

I fully agree we should support the `baseApiPath` giving users the ablity to modify where they place the API router.

However, I do have some questions regarding the `basePath` option:
* Would this suggest that the Faust app is deployed to a site (say `example.com`) in a subdir? So `example.com/blog`? 

If so, next.js does have a config options for this, [`basePath`](https://nextjs.org/docs/api-reference/next.config.js/basepath), which perhaps we can look into supporting instead of creating an additional option oursevles.

So instead of creating `basePath` in Faust config, the user would specify the `basePath` in their `next.config.js`, and we could use that to properly route requests. Thoughts?

riderjensen:
@blakewilson that is a great idea, thank you for the suggestion, it definitely makes more sense to just use the next.js config values. I updated some files, technically my implementation would still allow a user to set their `basePath` in the config however, it defaults to using next's configuration and the docs only mention using next's config and not faust's config. Do you think this is an acceptable solution? Or should I remove the `basePath` completely from the `FaustConfig` object and just import the next config in the files where we need it?

riderjensen:
@blakewilson did you have a chance to revisit the changes? I am sure you are busy so no rush, just wanted to ping you again to keep you aware

riderjensen:
I've added a change to no longer reference the next config as instead use environment variables. This is recommended from next.js [here](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration): 

```
Note: [Runtime Configuration] is considered legacy and does not work with Automatic Static Optimization, Output File Tracing, or React Server Components. Please use environment variables instead in order to avoid initialization overhead.
```

I also updated the documentation I wrote

riderjensen:
Ive updated the MR to remove the changes around `baseApiPath` and moved the focus of this MR down to just the change for `basePath` in order to fix the bug around endless refreshing.

blakewilson:
Thanks @riderjensen and apologies for the late response. This looks good, let me run it through a few test cases and I'll report back.

austintreneff:
Hey @blakewilson any update here?

internal/faustjs.org/docs/faustwp/settings.mdx
```diff
@@ -9,7 +9,24 @@ allows you to customize the behavior of the plugin by configuring certain featur
...
 
-The following options are available to enable/disabled based on your requirements:
+The following options are available to customize based on your requirements:
+
+### Modifying Base Path
+
+You can change the base path of the Front-end site URL by including it in your `Front-end site URL` value.
+For example, if your front-end site was hosted at `http://example.com/blog`, you would change the `Front-end site URL` value to `http://example.com/blog`.
+You also need to do two additional things to get the base path working. First, add an environment variable called `NEXT_PUBLIC_BASE_PATH` in yoru `.env` file and set it to `/blog`.
+Then reference your env value in your `next.config.js` for authenticating post previews to continue working.
```
blakewilson, line 19:
```suggestion
You also need to add your base path value in your `next.config.js` for authenticating post previews to continue working.
```

internal/faustjs.org/docs/faustwp/settings.mdx
```diff
@@ -9,7 +9,24 @@ allows you to customize the behavior of the plugin by configuring certain featur
...
+
+```js title="next.config.js" {15}
+const { withFaust } = require('faustwp-core')
+
+module.exports = withFaust({
+  ...
+  basePath: process.env.NEXT_PUBLIC_BASE_PATH,
```
blakewilson, line 26:
```suggestion
  basePath: '/blog'
```

blakewilson (changes requested):
Thanks for your patience on this @riderjensen. This all looks great except for the `NEXT_PUBLIC_BASE_PATH` env var. We are currently trying to not add any more env vars that we'll need to support long term.

Regardless, this feature will still be supported by passing in the `basePath` value via `faust.config.js`:

```js
module.exports = withFaust({
   basePath: '/blog'
})
```

Once these changes have been made, we can get this merged in!

riderjensen:
Hey @blakewilson thanks for suggesting those changes! I included them in the PR, I see there is a failing test but I dont know if that is caused by my changes. Let me know if there is anything else I can do to get this merged in!

blakewilson:
@riderjensen All tests have passed, it seems those failures are since you are outside of the org and don't have proper permissions. This PR looks great, thanks again for your patience and work on this!

blakewilson (approved):

PR #1292, opened 2023-02-24
Provide skeleton component for individual toolbar nodes

josephfusco:
## Tasks

## Description

These changes introduce a new skeleton component that replaces internal "Loading..." fragments. This component has also been made available for users to use in custom nodes.

## Testing

1. Checkout this branch: `MERL-698-skeleton-loader`
2. Run `npm run build -w @faustwp/core`
4. Drop [this Faust plugin](https://gist.github.com/josephfusco/b781839e8c33dc729b64f94eaa8a57c3) into your test site
5. Register the plugin from step 3 in `faust.config.js`
    ```js
      experimentalPlugins: [new CustomToolbar()],
    ```
6. Observe 2 toolbar nodes with a pulsing background.
7. Observe 2 toolbar nodes with a static background when a user's system setting is set to [prefer reduced motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion).

## Screenshots

no motion preference
prefers reduced motion

### DIFF
--- a/packages/faustwp-core/src/components/Toolbar/ToolbarNodeSkeleton.tsx
+++ b/packages/faustwp-core/src/components/Toolbar/ToolbarNodeSkeleton.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+/**
+ * Skeleton component for a Toolbar Node.
+ */
+export function ToolbarNodeSkeleton() {
+  return (
+    <div className='toolbar-node-skeleton-wrapper'>
+      <div aria-hidden="true" className='toolbar-node-skeleton' />
+    </div>
+  );
+}
--- a/packages/faustwp-core/src/components/Toolbar/ToolbarNodeSkeleton.tsx
+++ b/packages/faustwp-core/src/components/Toolbar/ToolbarNodeSkeleton.tsx
@@ -5,8 +5,8 @@ import React from 'react';
  */
 export function ToolbarNodeSkeleton() {
   return (
-    <div className='toolbar-node-skeleton-wrapper'>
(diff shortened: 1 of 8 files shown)


packages/faustwp-core/src/components/Toolbar/ToolbarNodeSkeleton.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+/**
+ * Skeleton component for a Toolbar Node.
+ */
+export function ToolbarNodeSkeleton() {
+  return (
+    <div className='toolbar-node-skeleton-wrapper'>
```
blakewilson: Looks like there is some linting errors here

blakewilson (commented):
There are a few linting issues, but besides that looks good. Nice work @josephfusco!

blakewilson:
@josephfusco Do you think users will know what `ToolbarNodeSkeleton` does? Should we call this something more apparent like `ToolbarNodeLoading`?

josephfusco:
@blakewilson Good question!

I feel that indicating it's a skeleton is important for semantics. I understand this term may not be common knowledge. Maybe renaming to `ToolbarNodeSkeletonLoader` offers improved readability for those folks?

Other libraries using the term skeleton

- https://mui.com/material-ui/react-skeleton/
- https://chakra-ui.com/docs/components/skeleton

blakewilson:
@josephfusco Awesome! Just wanted to make sure that was the norm in UI libs, looks like it is.

blakewilson (approved):
LGTM!

PR #1296, opened 2023-02-27
Export everything under the main entry point

josephfusco:
These changes fix an issue introduced in `@faustwp/core@0.2.5` where modules were not found.


blakewilson (approved):
LGTM

PR #1318, opened 2023-03-08
Have Faust CLI verify if WPGraphQL is active

josephfusco:
## Tasks

## Description

These changes handle verifying if the WPGraphQL plugin is active in order to provide more helpful info to a Faust user. Prior to these changes, if WPGraphQL was unavailable, a [Next.js pre-render error](https://nextjs.org/docs/messages/prerender-error) was thrown that wasn't too helpful.

These changes also refactor the internal structure of this package as the utils/ directory was getting a little challenging to visually search through.

These changes also setup Jest tests within the `@faustwp/cli` package. 🎉 

## Testing

1. Clone this branch: `cli-health-check`
2. Update the faustwp example project to use a test WordPress site
3. Deactivate WPGraphQL from your connected WordPress site
4. Run `npm run build -w @faustwp/getting-started-example`
5. Observe that the build command will warn user and is non-blocking (see screenshot below - note that the error is now a warning)
6. Re-activate WPGraphQL
7. Run `npm run build -w @faustwp/getting-started-example`
8. Observe that the build command was successful

## Screenshots

Screenshot 2023-03-08 at 2 01 26 PM


.changeset/thick-lizards-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/cli': patch
```
theodesp, line 2: since this is a new feature it should be a `minor` change

packages/faustwp-cli/utils/getGraphqlEndpoint.ts
```diff
@@ -0,0 +1,10 @@
+import { getWpUrl } from './getWpUrl.js';
+
+/**
+ * Returns the GraphQL Endpoint.
+ */
+export function getGraphqlEndpoint() {
+  const wpUrl = getWpUrl();
+
+  return `${wpUrl}/graphql`;
```
theodesp, line 9: This could be problematic if the users change that using the `https://www.wpgraphql.com/filters/graphql_endpoint` filter

packages/faustwp-cli/utils/getGraphqlEndpoint.ts
```diff
@@ -0,0 +1,10 @@
+import { getWpUrl } from './getWpUrl.js';
+
+/**
+ * Returns the GraphQL Endpoint.
+ */
+export function getGraphqlEndpoint() {
```
theodesp, line 6: We should definitely need unit tests here to cover edge cases.

packages/faustwp-cli/utils/getWpUrl.ts
```diff
@@ -0,0 +1,14 @@
+import trim from 'lodash/trim.js';
+
+/**
+ * Returns the WordPress URL from the project's env.
+ */
+export function getWpUrl(path = ''): string {
+  const wpUrl = trim(process.env.NEXT_PUBLIC_WORDPRESS_URL, '/');
```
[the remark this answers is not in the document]
blakewilson: This can get tricky because we have filters in the plugin system that users may rely on to modify their `wpUrl()` but they won't be applied here in the CLI.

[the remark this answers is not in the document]
josephfusco: 💯 agree that filters should work across all package (blocks, core & cli) and one that I would like to solve, however that issue existed prior to these changes, and think it should be dealt with in a separate PR.

The utility functions here were merely a means to centrally locate some repetitive code within the cli package.

Merging the core & cli packages could get us closer to that reality, but outside the scope of this PR.

packages/faustwp-cli/utils/getGraphqlEndpoint.ts
josephfusco: Thanks @theodesp, I absolutely agree. [See my comment here](https://github.com/wpengine/faustjs/pull/1318#discussion_r1132522872).

packages/faustwp-cli/utils/getWpUrl.ts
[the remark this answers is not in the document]
blakewilson: The challenge is that the introduction of this PR could cause a breaking change. Say for instance a user has their WPGraphQL endpoint as `/gql` instead of the default `/graphql`, and they are using the `graphqlEndpoint` Faust filter to modify that during runtime. With the introduction of this PR without the filter, their build would fail as the `/graphql` endpoint does not exist for them.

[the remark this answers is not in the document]
josephfusco: @blakewilson 💯 - [this is already an issue with the current version](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-cli/utils/generatePossibleTypes.ts#L21). This PR is just centralizing that code.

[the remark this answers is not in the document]
blakewilson: My concern is that while it is an issue in the current version, it is contained to just the `faust generatePossilbeTypes` script. So a user can still use Faust (they would have to generate types their own way if they stray from the default in the example project), but still usable. By making the changes in this PR, they are now blocked from even running `faust dev`, `faust build`, `faust start`, etc. And their Faust app breaks entirely until they adopt the `/graphql` endpoint naming convention.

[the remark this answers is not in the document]
josephfusco: @theodesp @blakewilson Updated to throw a non-blocking warning message.

packages/faustwp-cli/tsconfig.json
```diff
@@ -1,12 +1,15 @@
...
-  "exclude": ["dist", "./*.d.ts"]
+	"compilerOptions": {
+		"strict": true,
+		"outDir": "dist",
+		"module": "ES2015",
+		"target": "ES2017",
+		"esModuleInterop": true,
+		"moduleResolution": "node",
+		"declaration": true
+	},
```
blakewilson, line 10:
```suggestion
	},
	"include": ["src"],
```

Very important we include the newly created `src` directory in the `tsconfig.json`. Without it, our dist directory includes `src` (i.e. `dist/src/index.js`) and the build will break.

Screenshot 2023-03-22 at 2 09 49 PM

packages/faustwp-cli/tests/healthCheck/verifyGraphQLEndpoint.test.ts
blakewilson, line 1: Nice! Finally we've got some tests for the CLI!

blakewilson:
Nice, I really like the organization of the new `src` directory.

blakewilson:
Might need some other testers to check this out. I get the following whenever trying to run the CLI after pulling the PR:

Screenshot 2023-03-22 at 2 14 51 PM

This could def be related to my local machine. But it would be helpful if others could prove/disprove this.

packages/faustwp-cli/src/healthCheck/validateFaustEnvVars.ts
```diff
@@ -20,8 +20,12 @@ export const validateFaustEnvVars = () => {
     process.env.NEXT_PUBLIC_WORDPRESS_URL.startsWith('http://') &&
     getWpSecret()
   ) {
+    warnLog('Your WordPress site is not running on https!');
     warnLog(
-      'Your WordPress site is not running on https! This is a security concern as all traffic with your secret key is in plain text. Please make sure your production Faust app runs with a WordPress instance on https!',
+      'This is a security concern as all traffic with your secret key is in plain text.',
+    );
+    warnLog(
+      'Please make sure your production Faust app runs with a WordPress instance on https!',
```
josephfusco, line 28: Shorten line length to improve readability

packages/faustwp-core/src/lib/getGraphqlEndpoint.ts
```diff
@@ -3,7 +3,7 @@ import { getWpUrl } from './getWpUrl.js';
 
 export function getGraphqlEndpoint(): string {
   const wpUrl = getWpUrl();
-  let graphqlEndpoint = `${wpUrl}/graphql`;
+  let graphqlEndpoint = `${wpUrl}/index.php?graphql`;
```
josephfusco, line 6: This PR makes the existing GraphQL filter useless so we may want to follow up with a deprecation of https://faustjs.org/docs/plugin-system/filters#graphqlendpoint

packages/faustwp-cli/src/utils/getWpUrl.ts
```diff
@@ -0,0 +1,14 @@
+import trim from 'lodash/trim.js';
```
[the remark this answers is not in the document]
josephfusco: @blakewilson See https://github.com/wpengine/faustjs/pull/1318#issuecomment-1501901403 🎉

.changeset/thick-lizards-smoke.md
blakewilson, line 1: We should also create a changeset for `@faustwp/core` since we updated the `/graphql` endpoint

packages/faustwp-cli/src/healthCheck/verifyGraphQLEndpoint.ts
```diff
@@ -0,0 +1,44 @@
...
+      };
+    } = await response.json();
+
+    // eslint-disable-next-line no-underscore-dangle
+    if (json.data.__typename) {
+      infoLog('Discovered WPGraphQL endpoint!');
+      return true;
+    }
+
+    return process.exit(0);
```
blakewilson, line 35: Technically this last `process.exit(0)` can happen without throwing the error messages below. If the response's payload is json but doesn't include `data.__typename`, it will just quit the execution without showing the error logs in the below catch method. Maybe instead if the `__typename` is not found, we throw an error, forcing it to be caught in the block below. For example:

```suggestion
    if (!json.data.__typename) {
      throw new Error(`GraphQL request for didn't include "json.data.__typename"`)
    }
```

Additionally, I'm wondering if we need a log to call out that the WPGraphQL endpoint was found. I think by default we should assume it is. Maybe if anything we include this as a debug message with the endpoint included for debug purposes.

blakewilson:
Also wanted to confirm with @jasonbahl, if we change our calls under the hood from `/graphql` to `index.php?graphql` we should see no changes correct? Is there any instances where `index.php?graphql` wouldn't work?

jasonbahl:
@blakewilson ya, the "GraphQL Endoint" setting here lets users change this value. 

if pretty permalinks are enabled it's `/$whateverThisValueIs` and if not it's `index.php?$whateverThisValueIs`. 

CleanShot 2023-04-11 at 13 15 16

Here' i've changed it to "kittens" so my GraphQL endpoint can be accessed at `index.php?kittens` or `/kittens`

CleanShot 2023-04-11 at 13 17 41

And `/graphql` and `index.php?graphql` 404 now.

jasonbahl:
@blakewilson ya, I think in the real early days (before the name Faust existed I think) instead of defining an .env var to point to your WordPress url and `/graphql` or `?graphql` being hard-coded, it was an .env var that pointed at your WPGraphQL endpoint, whatever it may be, and from there all info was gathered. 

I'd probably still recommend an .env var to the graphql endpoint vs hard-coding `/graphql` or `?graphql`

PR #1355, opened 2023-03-27
MERL-693: flatListToHierarchical rewrite

matthewguywright:
## Description
Move flatListToHierarchical from projects into faustwp/core.


internal/faustjs.org/docs/gutenberg/wp-graphql-content-blocks.mdx
```diff
@@ -253,7 +253,7 @@ This way you can reconstruct the block tree as before and pass it on to the `Wor
 
 ```js
 import { WordPressBlocksViewer } from '@faustwp/blocks';
-import flatListToHierarchical from '../utilities/flatListToHierarchical';
+import { flatListToHierarchical } from '@faustwp/core/src';
```
josephfusco, line 256: We will want users to be able to import this function directly from @faustwp/core (without the need for /src)

packages/faustwp-core/tests/utils/flatListToHierarchical.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    id: '5',
+    name: 'abc',
+    parentId: '',
+    children: [{ id: '3', name: 'abc', parentId: '5', children: [] }],
+  },
+];
+
+describe('utils/flatListToHierarchical', () => {
+  test('returns a hierarchical list from items array', () => {
+    //@ts-ignore
```
josephfusco, line 43: Is this ignore needed?

packages/faustwp-core/src/utils/flatListToHierarchical.ts
```diff
@@ -0,0 +1,44 @@
...
+  childrenKey?: string;
+}
+
+/**
+ * Converts a flat list to hierarchical.
+ * @param data The node items.
+ * @param param1 The node parameters.
+ * @returns Node Array
+ */
+export default function flatListToHierarchical(
```
josephfusco, line 19: We should use a named export here to match the code style of this project.

examples/next/faustwp-getting-started/components/NavigationMenu/NavigationMenu.js
```diff
@@ -1,9 +1,9 @@
 import classNames from 'classnames/bind';
 import { gql } from '@apollo/client';
 import Link from 'next/link';
-import flatListToHierarchical from '../../utilities/flatListToHierarchical';
 import styles from './NavigationMenu.module.scss';
 import stylesFromWP from './NavigationMenuClassesFromWP.module.scss';
+import { flatListToHierarchical } from '@faustwp/core/src';
```
josephfusco, line 6: We will want users to be able to import this function directly from @faustwp/core (without the need for /src)

packages/faustwp-core/tests/utils/flatListToHierarchical.test.ts
```diff
@@ -0,0 +1,46 @@
+import flatListToHierarchical from '../../src/utils/flatListToHierarchical';
```
josephfusco, line 1: We should import the utility function from packages/faustwp-core/src/utils/index.ts since it's available.

josephfusco (changes requested):
Hey @matthewguywright, left some feedback for these changes.

I am working on a PR for those linting errors being thrown.

packages/faustwp-core/tests/utils/flatListToHierarchical.test.ts
matthewguywright: This is the current pattern in place for the tests. They are using relative paths.

theodesp:
@matthewguywright you want to do a rebase since the build is failing here.

packages/faustwp-core/src/index.ts
```diff
@@ -41,8 +41,14 @@ import {
...
   ToolbarSubmenuWrapper,
 } from './components/Toolbar/index.js';
+import {
+  flatListToHierarchical,
+  ListNode,
+} from './utils/flatListToHierarchical.js';
 
 export {
+  flatListToHierarchical,
+  ListNode,
```
blakewilson, line 51: Do we need to expose `ListNode` here?

blakewilson:
@matthewguywright We'll also want a changeset

packages/faustwp-core/src/index.ts
matthewguywright: Yes

matthewguywright: Yes. We had to make it publicly exposed through /core. Previously it had Node and that was clashing with some other namespace so we created and exposed this.

blakewilson: @matthewguywright What is it used for? Isn't it just a type for an internal param?

If we expose it in the main export we'll need to support it indefinitely since it'll become apart of our public API.

blakewilson (approved):
LGTM 👍

PR #1361, opened 2023-03-29
Pass data and variables in next helpers

blakewilson:
## Tasks

## Description

This PR adds the Next.js file based page's query response as `props.data` in both `getNextServerSideProps` and `getNextStaticProps`

Additionally, adds a `FaustPage` type much like `FaustTemplate` to help properly type Next.js pages that use Faust helpers.

## Testing

1. Checkout this PR
2. Build core `npm run build -w @faustwp/core`
3. Run the example project `npm run dev -w @faustwp/getting-started-example`
4. In the `example.js` page, console.log the `props`
5. Notice that there is now `props.data` which includes the `Page.query`'s data, and `__PAGE_VARIABLES__` prop, which includes the computed variables (this value should not really be used, but we include it just in case to be compatible with the Faust Templates)
Screenshot 2023-03-29 at 3 41 24 PM

### DIFF
--- a/.changeset/lucky-islands-deny.md
+++ b/.changeset/lucky-islands-deny.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/core': patch
+---
+
+Return `data` in props from Next.js pages that use the `getNextStaticProps`/`getNextServerSideProps` Faust helper functions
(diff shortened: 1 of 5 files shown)


theodesp (approved):
LGTM!

PR #1352, opened 2023-03-24
Track Faust bundle sizes

josephfusco:
## Tasks

- [X] I have signed a [Contributor License Agreement (CLA)](https://github.com/wpengine/faustjs#contributor-license-agreement) with WP Engine.

## Description

These changes add a new workflow that will perform a bundle analysis using [@next/bundle-analyzer](https://www.npmjs.com/package/@next/bundle-analyzer) and compare it to the bundle analysis in the base branch.

## Testing

The bundle analysis for **this PR is expected to fail** this first time
due to there being no previously ran bundle analysis in `canary`.

You can view this [testing pull request](https://github.com/wpengine/faustjs/pull/1365) to see a real bundle size change report.

**No need for a changeset** since this is workflow related.

PR #1365, opened 2023-04-03
Force bundle size increase

josephfusco:


.github/workflows/nextjs-bundle-analysis.yml
josephfusco, line 1: Ignore the changes in this file as they have since been pulled into the [feature PR](https://github.com/wpengine/faustjs/pull/1352)

Issue #1344, opened 2023-03-21
Request: Default block improvements


justlevine:
The current provided default block wraps everything in an unnecessary `<div>`. While native React requires us to only `dangerouslySetInnerHTML` on a valid HTML tag, it provides an extra barrier for getting Faust set up. I recommend the following changes:

1. Provide a WP filter to let a user provide their own default block. That way a user can choose to wrap them in `span` or use an external library to ditch the wrapper.

2. Consider using `html-react-parser` (or another library) for the default block. Even though its an external dependency, I believe the benefits of both dropping the router, and using parser options to replace html tags with components (e.g. next/image) outweigh that concern. (And of course, if so, please make the parser options filterable as well.

theodesp:
@justlevine We have a ticket in our backlog to configure this. I will prioritise for next sprint.

The API would be a bit simpler though:

* There will be a configuration to hook into the resolve block template function so that the user can supply its both their ow `DefaultBlock` or the logic to match the templates.
* Not sure about the html-react-parser since if you have the option to customize the `DefaultBlock` you can also bring your own parser like html-react-parser.

blakewilson:
Closed via #1368

PR #1368, opened 2023-04-04
Add support for modifying `resolveBlockTemplate` logic and the fallback (default) block

blakewilson:
## Tasks

## Description

This PR introduces a way to modify the `resolveBlockTemplate` logic (through the `faustBlocksResolveBlockTemplate` filter) , as well as a way to modify the fallback (default) block for `WordPressBlocksViewer`.

The `faustBlocksFallbackBlock` filter allows you to modify all default fallback blocks for `WordPressBlocksViewer`, but the component also accepts a `FallbackBlock` prop if you want to have a different fallback block for a specific `WordPressBlocksViewer`.

### DIFF
--- a/.changeset/fifty-ducks-trade.md
+++ b/.changeset/fifty-ducks-trade.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/core': patch
+---
+
+Export hooks from Faust core
--- a/.changeset/fifty-ducks-trade.md
+++ b/.changeset/fifty-ducks-trade.md
@@ -1,5 +1,0 @@
----
-'@faustwp/core': patch
----
-
-Export hooks from Faust core
--- a/packages/blocks/src/components/WordPressBlocksViewer.tsx
+++ b/packages/blocks/src/components/WordPressBlocksViewer.tsx
@@ -23,6 +23,7 @@ export function BlockDataProvider(
 
 export interface WordpressBlocksViewerProps {
   blocks: Array<ContentBlock | null>;
+  FallbackBlock: React.FC<ContentBlock>;
 }
 
 export interface ContentBlock {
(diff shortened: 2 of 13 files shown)


.changeset/fifty-ducks-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/core': patch
+---
+
+Export hooks from Faust core
```
theodesp, line 5: This should be a minor change since we are introducing something new.

blakewilson: @theodesp But there is no breaking changes here. Since we're still 0.x.x, I believe we're still using `patch`

packages/blocks/src/components/WordPressBlocksViewer.tsx
```diff
@@ -22,7 +22,26 @@ export function BlockDataProvider(
...
+   * const MyFallbackComponent = ({renderedHtml}) => {
+   *   return (
+   *     <span dangerouslySetInnerHTML={{__html: renderedHtml}} />
+   *   )
+   * }
+   *
+   * <WordPressBlocksViewer FallbackBlock={MyFallbackComponent} />
+   * ```
+   */
+  FallbackBlock: React.FC<ContentBlock>;
```
theodesp, line 44: Should use `camelCase` for component props here. it's a convention adopted by the official React documentation.

blakewilson:
@justlevine Does this satisfy #1344?

justlevine:
@blakewilson this is perfect. Especially love the flexibility of being able to use a filter or to pass it as a prop.

Plugin system is also a great alternative to bundling a parser in faust for a better `DefaultBlock` experience - and I'm sure _somebody_ will release one - so yeah #1344 can imo be closed in its entirety.

theodesp (approved):
LGTM!

PR #1387, opened 2023-04-18
Updates changesets CLI version to latest

blakewilson:
## Tasks

## Description

This should handle the creation of private package git tags correctly.

## Testing

Tested locally with this latest version of changesets in the Faust.js repo and the private package git tags were created as expected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "@faustjs/root",
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Looks good!

This has me thinking that we don't necessarily need to include changesets as a dependency if use `npx changeset` instead of `npm run changeset`.

Would just require updating the contributing docs.

PR #1390, opened 2023-04-19
Deploy the WordPress plugin after a successful changesets publish

blakewilson:
## Tasks

## Description

This PR deploys the WordPress plugin after changesets successfully "publishes" the `@faustwp/wordpress-plugin` package (The WordPress plugin). In the context of changesets, a successful publish is a new tag for a private package, or a publish to NPM if it's an npm package).

Currently, changesets creates the tag just fine for the WordPress plugin, but the action to deploy to WordPress is not being ran. That's because GitHub Actions can not invoke other actions to run:
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

So, instead of calling another action, this PR checks if the plugin was versioned during the "release packages" action, and if so, deploys it to wordpress.org


josephfusco (approved):
This is going to be a joy to have working!

Issue #1379, opened 2023-04-07
Bug: Error on client-side navigation to SSR page


JEverhart383:
In trying to create a multilingual example Faust site, I turned on SSR for [index.js](https://github.com/JEverhart383/faust-intl/blob/main/pages/index.js) using `getServerSideProps` so that Next.js would automatically detect the browser locale on the home page. That works great when the page first loads, but from there, if I click around the site via client navigation, and return to the SSR homepage, I get a 404 error. In digging into the code, it looks like this is because Next.js requests the JSON chunk for that page, which doesn't exist since it's supposed to be SSR.
![Screen Shot 2023-04-07 at 10 05 51 AM](https://user-images.githubusercontent.com/7818951/230622260-a03cb0a0-f142-4c20-acae-a2c26c17667a.png)

@jordanmaslyn was able to help troubleshoot and we narrowed it down to [this line of code](https://github.com/wpengine/faustjs/blob/cce2828214280a0b525dec08870fea7da64899da/packages/faustwp-core/src/getWordPressProps.tsx#L70). In this scenario, the `ctx.req.url` returns the Next chunk that doesn't exist, while `ctx.resolvedUrl` contains the correct route. Making this change in the local project fixed the issue.

## Applicable Versions

- `@faustjs/core` version:
- `@faustjs/react` version:
- `@faustjs/next` version:
- WordPress version:

## Steps To Reproduce

1. Open the [test site](https://hjlfagvo731xqt0ussridhbz9.js.wpenginepowered.com/)
2. Navigate to any page besides the homepage
3. Navigate to the homepage via main site nav link
4. See 404 regardless of locale

Link to code example:

https://github.com/JEverhart383/faust-intl/blob/main/pages/index.js

## The current behavior

Client-side navigations to SSR pages generate a 404 when using `getWordPressProps`

## The expected behavior
My page should be SSR'd

theodesp:
Hey @JEverhart383  Thanks for the suggestion I was able to verify.

It seems that when navigating from a static page to getServerSiteProps page then the router will perform an API request to resolve the page data

https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#when-does-getserversideprops-run

this is why you get 404 on `en.json` since it's not a valid route for the seed query. I will push a PR for this.

PR #1395, opened 2023-04-24
Bug Fix: Error on client-side navigation to SSR page

theodesp:
## Tasks

## Description
Bug Fix for https://github.com/wpengine/faustjs/issues/1379

## Testing
Follow the instructions of the original issue to test a site using `getServerSideProps` on the index page.

### DIFF
--- a/packages/faustwp-core/src/getWordPressProps.tsx
+++ b/packages/faustwp-core/src/getWordPressProps.tsx
@@ -67,7 +67,7 @@ export async function getWordPressProps(
       resolvedUrl = '/';
     }
   } else {
-    resolvedUrl = ctx.req.url;
+    resolvedUrl = ctx.resolvedUrl;
     ctx.res.setHeader('x-using', 'faust');
   }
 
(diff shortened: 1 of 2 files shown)


packages/faustwp-core/src/getWordPressProps.tsx
```diff
@@ -67,7 +67,7 @@ export async function getWordPressProps(
       resolvedUrl = '/';
     }
   } else {
-    resolvedUrl = ctx.req.url;
+    resolvedUrl = ctx.resolvedUrl;
```
blakewilson, line 70: Wouldn't this set the `resolvedUrl` in SSR mode to `null`?

theodesp: I haven't seen null or undefined in SSR here but I've seen `ctx.req.url` to be undefined before.

theodesp:
```
resolvedUrl: A normalized version of the request URL that strips the _next/data prefix for client transitions and includes original query values.
```
https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter

blakewilson (approved):

PR #1463, opened 2023-06-15
fix: GetSeedNode `id` and `uri` GraphQL params are required.

justlevine:
## Tasks

## Description
This PR fixes a bug in the updated `GetSeedNode()` GraphQL query, where the `$id` and `$uri` parameters are given the wrong type (they must be non-null).

While this doesn't trigger any errors on a regular GraphQL requests (since a default value means the parameter is non-null), it does throw a `GraphQLError` when used with WPGraphQL Smart Cache's persisted queries, as the default value gets stripped from the stored query.

**Note:** I was lazy and made this changes on the GH browser. I recommend a `squash` merge to cleanup the unnecessary individual commits. Similarly there's no changeset. This branch is open to commits from maintainers, so go wild.

### DIFF
--- a/internal/faustjs.org/docs/faustwp/seed-query.mdx
+++ b/internal/faustjs.org/docs/faustwp/seed-query.mdx
@@ -43,8 +43,8 @@ export interface SeedNode {
 
 export const SEED_QUERY = gql`
   query GetSeedNode(
-    $id: ID = 0
-    $uri: String = ""
+    $id: ID! = 0
+    $uri: String! = ""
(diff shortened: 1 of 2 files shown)


theodesp:
@justlevine Thanks we will take a look.

theodesp:
@justlevine can you make a changeset as well please?

justlevine:
@theodesp im AFK, but like I wrote in the PR, I left the branch open to commits. It's a nonbreaking change.

blakewilson (approved):

PR #1510, opened 2023-07-24
[POC] Decoupled Blocks

josephfusco:
## Tasks

## Description

This pull request demonstrates how the Faust.js framework could handle registering blocks that are located within the front-end application - "Decoupled Blocks"

- [X] New example project in order to demonstrate block registration
- [X] New CLI command to generate theme.json from WordPress
- [X] New CLI command to discover blocks and push to WordPress
- [X] New authenticated REST endpoint in `faustwp` to handle theme.json CLI command  
- [ ] New authenticated REST endpoint in `faustwp` to handle WP block creation CLI command

In addition, the new Next.js example (`./examples/next/faustwp-block-support`) takes advantage of the CSS variables provided by [WordPress global styles](https://fullsiteediting.com/lessons/global-styles/).


examples/next/faustwp-block-support/wp-templates/single.js
```diff
@@ -0,0 +1,115 @@
...
+        node {
+          name
+        }
+      }
+      editorBlocks(flat: false) {
+        name
+        __typename
+        renderedHtml
+        id: clientId
+        parentId: parentClientId
```
theodesp, line 88: That should contain the block fragments otherwise it won 't work.

examples/next/faustwp-block-support/wp-templates/page.js
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    page(id: $databaseId, idType: DATABASE_ID, asPreview: $asPreview) {
+      title
+      content
+      editorBlocks(flat: false) {
+        name
+        __typename
+        renderedHtml
+        id: clientId
+        parentId: parentClientId
```
theodesp, line 86: Block fragments are missing here.

examples/next/faustwp-block-support/styles/_base.scss
```diff
@@ -0,0 +1,59 @@
+// Base styles
+
```
theodesp, line 2: Ideally we don't want the base styles since it will interfere with the globalStyleSheet. So you might find the the styles are not in parity with the editor.

josephfusco:
@theodesp Missing fragments have been added to this pull request, however an error is now being thrown for conflicting types.

> ApolloError: Fields "attributes" conflict because subfields "width" conflict because they return conflicting types String and Float. Use different aliases on the fields to fetch both if this was intentional.

Thoughts?

plugins/faustwp/includes/blocks/functions.php
```diff
@@ -0,0 +1,34 @@
...
+	exit;
+}
+
+/**
+ * Returns the contents of theme.json if present.
+ *
+ * @return object The contents of theme.json.
+ */
+function get_theme_json() {
+	$file_path = get_template_directory() . '/theme.json';
```
justlevine, line 20: I believe we want to use the [processed theme.json data](https://developer.wordpress.org/reference/classes/wp_theme_json_resolver/), not just the default values in `theme.json`.

josephfusco: Thanks @justlevine that makes sense, I appreciate the feedback!

Would you say that [WP_Theme_JSON_Resolver::get_merged_data](https://developer.wordpress.org/reference/classes/wp_theme_json_resolver/get_merged_data/) would be the appropriate method in this case?

justlevine: Sure, if it makes sense to roll our own caching. Otherwise, probably [wp_get_global_settings()](https://developer.wordpress.org/reference/functions/wp_get_global_settings/).

josephfusco:
This POC started with us copying over the faustwp-getting-started example, however that example project contains a lot of components that aren't related to configuring block support.

I've opened a new PR with a much smaller example project that demonstrates how to configure block support https://github.com/wpengine/faustjs/pull/1565

I am closing this PR out in favor of breaking this POC up into more bite sized pieces, with the first one being the [block support example project](https://github.com/wpengine/faustjs/pull/1565).

PR #1511, opened 2023-07-25
MERL-792: Preview Links

matthewguywright:
## Tasks

## Description
Left WP preview link intact and added a faust headless link.

## Testing
Verify you see two preview links and one that is specific for headless if the urls aren't the same. If the urls are the same you should only see one link.


plugins/faustwp/includes/replacement/callbacks.php
```diff
@@ -97,94 +97,6 @@ function image_source_srcset_replacement( $sources ) {
...
-		$link = add_query_arg(
-			array(
-				$args,
-			),
-			$link
-		);
-	}
-
-	return $link;
-}
```
blakewilson, line 186: I think we will still need this code (in a separate function) that we can call to generate the headless preview link.

plugins/faustwp/includes/replacement/previewlinks.js
```diff
@@ -18,38 +18,31 @@ window.addEventListener('DOMContentLoaded', function () {
...
 
-              wp.data.dispatch('core/editor').autosave();
-            });
+      clearInterval(intervalId);
 
-            link.parentElement.insertBefore(copy, link);
-            link.style.display = 'none';
-          });
-        }, 100);
+      var faustPreviewLink = wpPreviewLink.clone();
```
theodesp, line 36:
```suggestion
      const faustPreviewLink = wpPreviewLink.clone();
```

plugins/faustwp/includes/replacement/previewlinks.js
```diff
@@ -18,38 +18,31 @@ window.addEventListener('DOMContentLoaded', function () {
...
+      faustPreviewLink[0].href = previewLink;
+      faustPreviewLink.html(
```
theodesp, line 38: html needs to be valid here. Wrap this with a `span` element.

theodesp:
@matthewguywright Unit and e2e tests are also failing and need to be updated.

PR #1528, opened 2023-08-09
Feat: registerFaustBlock helper

theodesp:
## Tasks

- [x ] I have signed a [Contributor License Agreement (CLA)](https://github.com/wpengine/faustjs#contributor-license-agreement) with WP Engine.

## Description

* Includes a new helper function `registerFaustBlock` that wraps `registerBlockType` function.
* Adds unit tests.


examples/next/getting-started/tsconfig.json
```diff
@@ -18,7 +18,8 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "baseUrl": "./src",
-    "incremental": true
+    "incremental": true,
+	"typeRoots": ["node_modules/@types"]
```
theodesp, line 22: Prevents type conflicts between different React Versions. So we need to specify explicitly the location of `typeRoots`.

packages/block-editor-utils/src/types/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare global {
+  let wp: unknown;
```
theodesp, line 2: Global `wp` version is available only within the WordPress installation context.

blakewilson (approved):
Nice @theodesp! Looks good. We may not need the changesets for `@faustwp/core` and `@faustjs/next` since it's just dev related, but I'll let you make the final call.

packages/faustwp-cli/src/telemetry/marshallTelemetryData.ts
```diff
@@ -1,5 +1,5 @@
 import fs from 'fs';
-import { platform } from 'node:process';
+import { platform } from 'process';
```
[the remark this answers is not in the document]
blakewilson: Just adding this link here for context, `node:process` and `process` resolve to the same thing:

https://stackoverflow.com/a/71473519

theodesp:
> Nice @theodesp! Looks good. We may not need the changesets for `@faustwp/core` and `@faustjs/next` since it's just dev related, but I'll let you make the final call.

I will add them just in case we need to do a release. It's a very small change anyway.

PR #1542, opened 2023-08-21
Improve toolbar accessibility

josephfusco:
## Tasks

- [X] If a code change, I have written testing instructions that the whole team & outside contributors can understand.
- [X] I have written and included a comprehensive [changeset](https://github.com/wpengine/faustjs/blob/canary/DEVELOPMENT.md#deployment) to properly document the changes I've made.

## Description

These changes improve the toolbar accessibility by allowing menus to be opened with the "enter" key. Once the menu is opened, a user should be able to tab through the menu items.

## Testing

1. Checkout branch: `MERL-941-toolbar-a11y`
2. `npm run build -w @faustwp/core`
3. `npm run dev -w @faustwp/getting-started-example`
4. Go to your connected WordPress site and preview a post to view/authenticate in Faust.
5. Tab over to the "My Account" dropdown on the right side of the Toolbar.
6. Observe that menu opens when "enter" or "space" is pressed, and you can tab through the menu while it's open.


packages/faustwp-core/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -3,19 +3,11 @@ import React, { PropsWithChildren } from 'react';
 type Props = PropsWithChildren<{
   href?: string;
   tabIndex?: number;
-  handleClick?: () => Promise<void>;
 }> & { [key: string]: unknown };
 
-export function ToolbarItem({ href, handleClick, children, ...props }: Props) {
```
blakewilson, line 9: By removing `handleClick`, will this effect any users that may be adding custom toolbar nodes?

Just want to make sure this isn't a breaking change.

josephfusco: Yes, this would affect anyone already using handleClick, as it's now relying on folks to pass in native attributes which should be caught be ...props.

An example of the new ToolbarItem can be seen here, where we are passing in native attributes to separate click events from keyboard events.

I can update the changelog to reflect that it is a breaking change.

blakewilson: @josephfusco Yeah I mean toolbar is still "experimental", so maybe no changeset bump, but we should call it out in the  changeset desc.

TeresaGobble:
Hi @josephfusco, this is awesome! I like the clear outline for each selected menu item a lot. I had one question on the behavior expected when pressing space to open the "My Account" dropdown on the right side of the Toolbar. It seems to also advance the scroll on the page. Is this unintended? 

https://github.com/wpengine/faustjs/assets/50935135/ad5ecec2-6579-4d76-9d40-f9bcaf58f035

josephfusco:
@TeresaGobble I can confirm the same behavior - thank you for raising this concern!

I took a look at how WordPress handles this same menu and "enter" is the only key that opens the menu. Given that, I feel it's safe to remove "space" match the behavior in WordPress core.

TeresaGobble (approved):
Awesome Joe, LGTM!

PR #1544, opened 2023-08-22
App Router: `getAuthClient` and `faustRouteHandler`

blakewilson:
## Tasks

## Description

This PR sets up the new API Router but for App Router, called `faustRouteHandler`, and creates the `getAuthClient` function. This is a function, when authenticated, will return an Apollo client with the access token configured as the `Authorization` bearer token. When unauthenticated, the client returns `null`. 

## Testing

**Additional tickets will be created for proper auth methods like login/logout. But for the scope of this ticket, we will set the tokens manually.**

1. `npm install` from the monorepo root
2. `npm run build` from the monorepo root to build the packages
3. `npm run dev -w @faustwp/app-router-example` to run the example project
4. Navigate to `/gated-content` and notice the "You must be authenticated" text.
5. Go to your WordPress instance and open the GraphiQL IDE
6. Use the following mutation to generate an authorization code (this code is valid for 60 seconds):
    ```graphql
    mutation MyMutation {
      generateAuthorizationCode(input: {username: "admin", password: "admin"}) {
        clientMutationId
        code
        error
      }
    }
    ```
7. Encode the `code` for URLs. You can do that here: https://www.urlencoder.org
8.  With the encoded authorization code, visit `localhost:3000/api/faust/token?code=xxxx`, you should see a response with access tokens and refresh tokens.
9. Visiting the above endpoint will set the valid refresh token in the cookie. You can double check it was set by opening your developer tools > storage > cookies.
10. Navigation back to http://localhost:3000/gated-content and notice your authenticated viewer name and your posts 

Video demonstration of testing instructions:

https://github.com/wpengine/faustjs/assets/5946219/b4f64d57-a5d5-4cab-aa5a-af3fa940d3f0


examples/next/app-router/app/[postSlug]/page.js
```diff
@@ -8,7 +8,7 @@ export default async function Page(props) {
   const postSlug = props.params.postSlug;
 
   // Depending on if isPreview or not use the auth client or regular client
-  let client = getClient();
+  let client = await getClient();
```
Comment (blakewilson) [inline examples/next/app-router/app/[postSlug]/page.js:11]: The `getClient` function has been made `async` for future proofing. Right now there is no need, but there may in the future and I'd rather do the change now before a lot of users are using it.

examples/next/app-router/app/api/faust/[route]/route.js
Comment (blakewilson) [inline examples/next/app-router/app/api/faust/[route]/route.js:1]: This is what our new implementation of the Faust API Router looks like. Next.js App Router API routes are now called Route Handlers. The `route.js` page exports an HTTP verb (like `GET` or `POST`) to handle that HTTP method.

https://nextjs.org/docs/app/building-your-application/routing/route-handlers

examples/next/app-router/app/gated-content/page.js
```diff
@@ -0,0 +1,38 @@
+import { gql } from '@apollo/client';
+import { getAuthClient } from '@faustwp/experimental-app-router';
+
+export default async function Page() {
+  const client = await getAuthClient();
```
blakewilson, line 5: When the `getAuthClient` is called and there is no refresh token, the client will return `null` so you can easily provide an unauthenticated state, like:

```js
 if (!client) {
     return <>You must be authenticated</>;
 }
```

packages/experimental-app-router/src/client.ts
```diff
@@ -31,12 +35,28 @@ export const { getClient } = registerApolloClient(() => {
...
     uri: getGraphqlEndpoint(),
   });
 
   // @todo Create hook for client and options.
 
+  // If the request is coming from the auth client, apply the auth link.
+  if (authenticated) {
+    linkChain = setContext(async (_, { headers }) => {
+      // get the authentication token from local storage if it exists
+      const token = await fetchAccessToken();
```
blakewilson, line 48: Fetch the access token whenever we call `getAuthClient`. With the caching system in place in Next.js 13 and App Router, this request can be cached, and have that cache evicted before the `accessToken` expires, since we receive the expiration time as well.

packages/experimental-app-router/src/server/routeHandler/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function PostFn(req: Request) {
+  // Intentionally left empty if/until we have POST endpoints.
+}
+
+export const faustRouteHandler = {
```
blakewilson, line 22: For more info on route handlers:

https://nextjs.org/docs/app/building-your-application/routing/route-handlers

examples/next/app-router/.env.local.sample
```diff
@@ -1,5 +1,8 @@
 # Your WordPress site URL
 NEXT_PUBLIC_WORDPRESS_URL=https://faustexample.wpengine.com
 
+# Your Faust site URL
+NEXT_PUBLIC_URL=http://localhost:3000
```
blakewilson, line 5: Using `fetch` in node does not allow relative URLs. Since we are using `fetch` now within RSCs, we need to know the site URL.

examples/next/app-router/app/gated-content/page.js
theodesp: By default this would be `null` on this page. How do we really get this to authenticate?

theodesp:
Bundle analysis on the js asserts shows that even with the App router approach we are still shipping the Apollo Client:

Screenshot 2023-08-23 at 12 34 35

Screenshot 2023-08-23 at 12 54 06

This is not good since we will have the same issue here with the current Faust.js version. Any ideas why?

examples/next/app-router/app/gated-content/page.js
blakewilson: I just updated the testing instructions related to this.

mindctrl (approved):
LGTM.

Noting that Faust auth is broken in Safari, due to the cookie not being set, and we have that ticketed in Jira as MERL-987. Was able to confirm working auth in Chrome using the instructions above.

blakewilson:
@mindctrl As a part of this PR, I dug into that bug as well. Turns out it's not a bug in Faust but a bug/expected behavior in Safari:

After some investigation, it seems this is due to the `Secure` cookie property we apply here:

https://github.com/wpengine/faustjs/blob/76f926398db91d2c9e4c0058c4d03b8e3f08d996/packages/faustwp-core/src/server/auth/token.ts#L48-L55

`Secure` can only be set on an HTTPS connection, so if your local development site is on HTTP, the cookie will not save and you’ll end up in an endless loop. Using a local development site on HTTPS fixes the issue.

Chrome and other browsers ignore this when on `localhost`, but it looks like Safari doesn't.

PR #1545, opened 2023-08-22
chore: update Composer deps for FaustWP plugin

mindctrl:
## Description

Updates Composer dependencies for FaustWP plugin.

## Testing
Existing tests should continue to pass.


josephfusco (approved):
Aside from the unrelated linting failure this looks good!

PR #1554, opened 2023-08-28
Add tests for App Router utils

blakewilson:
## Tasks

## Description

This PR adds tests for the `fetchAccessToken`, `faustRouteHandler`, and `tokenHandler` utils.

### DIFF
--- a/packages/experimental-app-router/src/server/routeHandler/tokenHandler.ts
+++ b/packages/experimental-app-router/src/server/routeHandler/tokenHandler.ts
@@ -19,7 +19,7 @@ export async function tokenHandler(req: Request) {
     }
 
     const { url } = req;
-    const code = new URL(url).searchParams.get('code');
+    const code = new URL(url).searchParams.get('code') ?? undefined;
 
     const cookieStore = cookies();
     const cookieName = `${getWpUrl()}-rt`;
@@ -80,7 +80,7 @@ export async function tokenHandler(req: Request) {
      * and expiration.
      */
 
-    const res = NextResponse.json(data, {
+    const res = new NextResponse(JSON.stringify(data), {
       status: 200,
     });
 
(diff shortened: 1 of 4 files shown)


packages/experimental-app-router/src/server/routeHandler/tokenHandler.ts
```diff
@@ -19,7 +19,7 @@ export async function tokenHandler(req: Request) {
     }
 
     const { url } = req;
-    const code = new URL(url).searchParams.get('code');
+    const code = new URL(url).searchParams.get('code') ?? undefined;
```
blakewilson, line 22: `searchParams.get()` returns `null` if there isn't a search param. In current Faust, we return `undefined`, so this change brings parity to the existing API router. The authorize endpoint in the plugin technically supports both:

https://github.com/wpengine/faustjs/blob/1e50244db87185b03aee3851dc95e052174d862a/plugins/faustwp/includes/rest/callbacks.php#L167-L172

So I figure there is no need to make a changeset of this since it doesn't affect users.

packages/experimental-app-router/src/server/routeHandler/tokenHandler.ts
```diff
@@ -80,7 +80,7 @@ export async function tokenHandler(req: Request) {
      * and expiration.
      */
 
-    const res = NextResponse.json(data, {
+    const res = new NextResponse(JSON.stringify(data), {
```
blakewilson, line 83: Was experiencing some weird testing behavior with `NextResponse.json`, it seems others are too:

https://github.com/vercel/next.js/issues/48524

It appears it's related to earlier Node versions, so instead of calling the `json` method, I instead opted to just stringify the JSON myself.

mindctrl (approved):

PR #1555, opened 2023-08-28
Remove unnecessary `P` TypeScript variable

blakewilson:
## Tasks

## Description

This PR removes the unnecessary `P` TypeScript variable in `registerFaustBlock`

### DIFF
--- a/packages/block-editor-utils/src/registerFaustBlock.ts
+++ b/packages/block-editor-utils/src/registerFaustBlock.ts
@@ -8,7 +8,7 @@ import DefaultSaveFn from './components/Save.js';
 import DefaultEditFn from './components/Edit.js';
 import { BlockFC, ConfigType } from './types/index.js';
 
-export interface RegisterFaustBlockMetadata<P, T extends Record<string, any>> {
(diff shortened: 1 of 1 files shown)


packages/block-editor-utils/src/registerFaustBlock.ts
```diff
@@ -8,7 +8,7 @@ import DefaultSaveFn from './components/Save.js';
 import DefaultEditFn from './components/Edit.js';
 import { BlockFC, ConfigType } from './types/index.js';
 
-export interface RegisterFaustBlockMetadata<P, T extends Record<string, any>> {
+export interface RegisterFaustBlockMetadata<T extends Record<string, any>> {
```
blakewilson, line 11: This is technically a breaking change for any users who have already typed their `T` value, i.e.:

```
const res = registerFaustBlock<PValue, TValue>(...args)
```

However, we don't even have this function documented or an actual approach to using this yet. Should we make this change? Or just ignore this linting issue in ESLint? 

Either way, this variable doesn't seem to make any sense so may cause confusion in the long term.

theodesp (approved):

Issue #1515, opened 2023-08-01
Bug: "Edit Post" node in the toolbar not working when previewing posts


renatonascalves:
## Applicable Versions

- `@faustjs/core` version: 1.0.3
- `@faustjs/react` version: Not used.
- `@faustjs/next` version: Not used.
- WordPress version: 6.2.2

I followed this "Get started with Faust" guide: https://faustjs.org/tutorial/get-started-with-faust

## Steps To Reproduce

https://github.com/wpengine/faustjs/assets/19148962/8e6e4539-4688-4c15-9765-1397322d6bde

Link to code example:

See this Github issue for a technical overview of the problem: https://github.com/wp-graphql/wp-graphql/issues/2876

## The expected behavior

I expected the "Edit Post" toolbar node to be available when previewing posts and the link should point me to the edit post screen of the post I'm previewing it.

renatonascalves:
There is one way Faust could handle this in the meantime. Getting the id in the params and swap for the one in the response. In that way, the edit post toolbar node would be available.

mindctrl:
@renatonascalves thank you for the report! Added to our backlog and we'll take a look at this as soon as possible.

mindctrl:
PR in progress here: https://github.com/wpengine/faustjs/pull/1556

PR #1556, opened 2023-08-29
MERL-1004: fixed missing edit button in the Faust toolbar when previewing a post

TeresaGobble:
## Tasks

## Description

This PR addresses the missing edit button in the Faust toolbar when previewing a post. 
Ticket: https://wpengine.atlassian.net/jira/software/c/projects/MERL/boards/1007?modal=detail&selectedIssue=MERL-1004

Closes https://github.com/wpengine/faustjs/issues/1515

## Testing
**Before testing:** 
Enable the experimental toolbar in your test site- see [Customizing the Toolbar - Faust.js™](https://faustjs.org/docs/guides/custom-toolbar) for instructions.
Enable public route redirects in WordPress:
wp-admin → settings → faust → Enable public route redirects :white_check_mark:

**Testing:**
Start your test site after switching to this branch. 
go to `wp-admin` and create a new post. Add content, and then click on the laptop icon on the upper righthand side of the editor. Go to `Preview in new tab.`
image

Note the edit button present in the Faust toolbar.

image

### DIFF
--- a/.changeset/funny-starfishes-melt.md
+++ b/.changeset/funny-starfishes-melt.md
@@ -1,0 +1,5 @@
+---
+"@faustwp/core": patch
+---
+
+Bug where review routes missing the "Edit Post" toolbar node fixed
--- a/.changeset/funny-starfishes-melt.md
+++ b/.changeset/funny-starfishes-melt.md
@@ -2,4 +2,4 @@
 "@faustwp/core": patch
 ---
 
-Bug where review routes missing the "Edit Post" toolbar node fixed
+Fixed bug where the "Edit Post" link was missing in the Faust Toolbar
--- a/packages/faustwp-core/src/components/Toolbar/nodes/Edit.tsx
+++ b/packages/faustwp-core/src/components/Toolbar/nodes/Edit.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useRouter } from 'next/router.js';
 import { SeedNode } from '../../../queries/seedQuery.js';
 import { getAdminUrl } from '../../../lib/getAdminUrl.js';
 import { ToolbarItem } from '../index.js';
@@ -8,17 +9,23 @@ type Props = {
 };
 
 export function Edit({ seedNode }: Props) {
-  if (
-    seedNode === undefined ||
-    seedNode?.isFrontPage ||
-    seedNode?.isPostsPage
-  ) {
+  const {
+    query: { p, typeName },
+  } = useRouter();
+
+  if (seedNode?.isFrontPage || seedNode?.isPostsPage) {
     return null;
   }
 
   // eslint-disable-next-line no-underscore-dangle
-  const postType = seedNode?.__typename || '';
-  const postId = seedNode?.databaseId || '';
+  let postType = seedNode?.__typename || '';
+  let postId = seedNode?.databaseId || '';
+
+  // This is a preview.
+  if (!postId) {
+    postId = p;
+    postType = typeName;
+  }
 
   const editPostUrl = getAdminUrl(`post.php?post=${postId}&action=edit`);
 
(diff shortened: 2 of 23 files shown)


packages/faustwp-core/src/components/Toolbar/nodes/Edit.tsx
```diff
@@ -8,17 +9,23 @@ type Props = {
...
-  const postType = seedNode?.__typename || '';
-  const postId = seedNode?.databaseId || '';
+  let postType = seedNode?.__typename || '';
+  let postId = seedNode?.databaseId || '';
+
+  // This is a preview.
+  if (!postId) {
+    postId = p as string;
+    postType = typeName as string;
+  }
```
blakewilson, line 28: This assumes that if `postId` does not exist, it's a preview, and that `p` and `typeName` do exist. However, that may not always be the case. it's possible to **not** have a `seedNode` and not have a `p`/`typeName` from the URL as well (in the case of CSR rendered pages).

What if we did something like:

```suggestion
  // eslint-disable-next-line no-underscore-dangle
  const postType = seedNode?.__typename ?? (typeName as string | undefined);
  const postId = seedNode?.databaseId ?? (p as string | undefined);

  if (!postId || !postType) {
    return null;
  }
```

blakewilson (changes requested):

.changeset/funny-starfishes-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@faustwp/core": patch
+---
+
+Bug where review routes missing the "Edit Post" toolbar node fixed
```
mindctrl, line 5:
```suggestion
Fixed bug where the "Edit Post" link was missing in the Faust Toolbar
```

mindctrl:
This is working well for me in Chrome. Noted internally that it does not work in Safari, but seems to be related to existing auth/cookie issue in Safari, and not a blocker for merge here.

Would be nice to have some tests to confirm the Edit node appears as expected to avoid regressions.

TeresaGobble:
> This is working well for me in Chrome. Noted internally that it does not work in Safari, but seems to be related to existing auth/cookie issue in Safari, and not a blocker for merge here.
> 
> Would be nice to have some tests to confirm the Edit node appears as expected to avoid regressions.

Thank you John! Still working on getting the tests to pass here for the packages on Node.js 16 and 18- the logic has changed a bit with this implementation. 👍

TeresaGobble:
Hi @blakewilson @mindctrl ! Thank you for reviewing this for me!

My test packages for Node.js 16 and 18 failed upon adding our regression tests for the Edit Post Node in previews. I've confirmed that they're passing locally, but in our GitHub actions the test is receiving [one assertion call](http://expected%20two%20assertions%20to%20be%20called%20but%20received%20one%20assertion%20call./) instead of two, and [expects 3 items but receives 4](https://github.com/wpengine/faustjs/actions/runs/6041182270/job/16393924705?pr=1556#step:6:399:~:text=251-,Expected%3A%203,Received%3A%204,-253). I've dug around but can't seem to find an explanation for the discrepancy. Maybe something is unexpected about the way GitHub is handling [expect.assertions(number)](https://jestjs.io/docs/expect) in this scenario?

theodesp:
> Hi @blakewilson and @mindctrl ! Thank you so much for reviewing this for me!
> 
> My test packages for Node.js 16 and 18 failed upon adding our regression tests for the Edit Post Node in previews. I've confirmed that they're passing locally, but in our GitHub actions the test is receiving [one assertion call](http://expected%20two%20assertions%20to%20be%20called%20but%20received%20one%20assertion%20call./) instead of two, and [expects 3 items but receives 4](https://github.com/wpengine/faustjs/actions/runs/6041182270/job/16393924705?pr=1556#step:6:399:~:text=251-,Expected%3A%203,Received%3A%204,-253). I've dug around but can't seem to find an explanation for the discrepancy. Maybe something is unexpected about the way GitHub is handling [expect.assertions(number)](https://jestjs.io/docs/expect) in this scenario? I've also rerun the failed checks, no change.

It might be that the assertions fail because of the missing  `FAUST_SECRET_KEY`:

https://github.com/wpengine/faustjs/actions/runs/6041182270/job/16393924705?pr=1556#step:6:513

blakewilson (approved):
Nice work @TeresaGobble! Thanks for adding regression tests as well.

PR #1594, opened 2023-10-02
MERL-803: overriding introspection

matthewguywright:
## Tasks

## Description
Adds support for authenticated introspection queries, so that users are not required to turn on "Enable Public Introspection" in WPGraphQL.

## Testing
- Set up a WP site with the Faust plugin, WPGraphQL, etc.
- Make sure WPGraphQL does **not** have "Enable Public Introspection" turned on.
- Set up a Faust example project, configured to work with the WP site above.
- Check out `canary` branch
- Run `npm run generate` in your Faust app. See error about needing to enable public introspection.
- Check out this PR branch. Ensure both your WP and Faust app are using this branch.
- Build the Faust app
- Run `npm run generate`. See success.
- Try it with an incorrect secret key to see updated error messaging.


mindctrl:
@matthewguywright can you fix the phpcs issues here?

mindctrl (dismissed):
Functionally this seems to be doing what is expected, based on what I know.

@matthewguywright could you add testing instructions to this?

Also left a previous comment about phpcs issues that need to be addressed.

plugins/faustwp/includes/graphql/callbacks.php
```diff
@@ -44,6 +45,28 @@ function register_templates_field() {
 	);
 }
 
+add_filter( 'graphql_get_setting_section_field_value', __NAMESPACE__ . '\\filter_introspection', 10, 5 );
+
+function filter_introspection( $value, $default_value, $option_name, $section_fields, $section_name ) {
```
mindctrl, line 50: We should probably add a unit test for this function

josephfusco, line 50: We should probably document this function.

Tip: When leveraging a WordPress filter, you can typically search the codebase where the filter is being implemented and copy those filter parameters.

In this case, I did a search [for the filter string](https://github.com/search?q=repo%3Awp-graphql%2Fwp-graphql%20graphql_get_setting_section_field_value&type=code) and [located it here](https://github.com/wp-graphql/wp-graphql/blob/develop/access-functions.php#L837C1-L845C5).

```suggestion
add_filter( 'graphql_get_setting_section_field_value', __NAMESPACE__ . '\\filter_introspection', 10, 5 );
/**
 * Update WPGraphQL's introspection setting's label in order inform a user of the exception for Faust.
 *
 * @param mixed  $value          The value of the field
 * @param mixed  $default_value  The default value if there is no value set
 * @param string $option_name    The name of the option
 * @param array  $section_fields The setting values within the section
 * @param string $section_name   The name of the section the setting belongs to
 */
function filter_introspection( $value, $default_value, $option_name, $section_fields, $section_name ) {
```

mindctrl: Done in https://github.com/wpengine/faustjs/pull/1594/commits/9e13dbb6040c0c4ff1241e5fed3516fde91022eb

mindctrl: Added in https://github.com/wpengine/faustjs/pull/1594/commits/ceb2c5d8e07709ec76336cf24b394d11d9e94904

blakewilson (approved):
@mindctrl Very nice implementation! This'll be a huge DX win.

PR #1603, opened 2023-10-11
MERL-1228: Fix issue with rewriting links to front end

### LINKED ISSUE #1569: Bug: Replacement of media library URLs in post content
## Applicable Versions

- WordPress plugin version: 1.0.3
- WordPress version: 6.3.1

## Steps To Reproduce

1. Add a link in any post content to an item in the media library (e.g. a PDF for download)
2. Fetch the content using GraphQL and see that the media URL has been replaced with the frontend URL (which is normally not a valid URL).

## The current behavior

`includes/replacement/callbacks.php` is replacing any instance of the `site_url` using `the_content` filter.

## The expected behavior

It should only replace URLs that belong to permalinks, where it might be an internal link and should be handled by the frontend routing. Replacing media URLs will normally break those links.

I've also attempted a fix here to exclude the uploads directory from the replacements: https://github.com/clicksuite/faustjs/commit/cb6dcecfa05ea031e624da54b440440cc78a6860

Another option could be to look for file extensions with a regex similarly to the graphql URL replacement.

Can submit a PR for this if that would be useful!

mindctrl:
## Tasks

## Description

Adjusts the URL replacement logic to avoid rewriting links to files in the uploads dir.

## Related Issue(s):
- Closes #1569

## Testing

Bug is described here: https://github.com/wpengine/faustjs/issues/1569

Ensure that links to posts, images, files, etc. are rewritten as expected. Try to break it!


plugins/faustwp/includes/replacement/callbacks.php
```diff
@@ -41,12 +41,17 @@ function content_replacement( $content ) {
...
+		$content    = preg_replace( "#href=\"{$site_url}(?!{$upload_dir})#", "href=\"{$replacement}", $content );
+	} else {
+		$content = str_replace( "href=\"{$site_url}", "href=\"{$replacement}", $content );
+	}
 
 	return str_replace( 'href="//', 'href="/', $content );
 }
 
-add_filter( 'the_content', __NAMESPACE__ . '\\image_source_replacement' );
+// add_filter( 'the_content', __NAMESPACE__ . '\\image_source_replacement' );
```
mindctrl, line 54: The linter is complaining about this. I left it commented out until we had some good testing on this. Will update and remove the whole function if we determine that's doable.

theodesp:
@mindctrl the `postcss` audit fail can be fixed by setting the version of `node_modules/postcss` to `8.4.31`. npm audit fix does not seem to change this so you need to edit manually:

```
"node_modules/postcss": {
      "version": "8.4.31",
      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.4.31.tgz",
      "integrity": "sha512-PS08Iboia9mts/2ygV3eLpY5ghnUcfLV/EXTOW1E2qYxJKGGBUtNjN76FYHnMs36RmARn41bC0AZmn+rR0OVpQ==",
      "dev": true,
```

theodesp (approved):
LGTM!

josephfusco (approved):
LGTM!

blakewilson:
I just retested with a PDF. To me, it seems the "Use the WordPress domain for media URLs in post content" option is backwards. When it is **enabled**, it uses the frontend URL, and when it's **disabled**, it uses the WordPress URL 🤔 


https://github.com/wpengine/faustjs/assets/5946219/cbac5d7e-ed89-44fb-b342-cc0aed82a1cb

matthewguywright (approved):

PR #1626, opened 2023-10-27
Support for multiple queries in a Faust template

blakewilson:
## Tasks

## Description

Provide a way to use multiple queries in a Faust template.

Introduces the new `Component.queries` property to execute multiple queries, and the `useFaustQuery` hook to retrieve the data.

## Testing

1. Checkout this repo
2. Install dependencies `npm install`
3. Build the packages `npm run build`
4. Start the example project `npm run dev -w @faustwp/getting-started-example`
5. Navigate to a single post, and notice it renders as expected with the two queries provided in `wp-templates/single.js`

### DIFF
--- a/packages/faustwp-core/src/getWordPressProps.tsx
+++ b/packages/faustwp-core/src/getWordPressProps.tsx
@@ -1,12 +1,14 @@
-import { GetServerSidePropsContext, GetStaticPropsContext } from 'next';
+import { print } from '@apollo/client/utilities';
 import type { DocumentNode } from 'graphql';
-import { SeedNode, SEED_QUERY } from './queries/seedQuery.js';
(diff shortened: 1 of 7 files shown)


packages/faustwp-core/src/getWordPressProps.tsx
```diff
@@ -141,6 +144,25 @@ export async function getWordPressProps(
     });
   }
 
+  let queries: { [key: string]: string } | null = null;
+  if (template.queries) {
+    const queryCalls = template.queries.map((query) => {
+      return client.query({
+        query,
+        variables: templateVariables,
```
theodesp, line 152: This is the tricky part. Each query may need different variables and not we are bundling all the potential variables on each requests. This could get out of hand soon.

Issue #1613, opened 2023-10-17
Bug: nodeByUri return null in category page


idmahbub:
## Applicable Versions
- "graphql": "^16.6.0",
-  "@apollo/client": "^3.6.6"
- "@faustwp/cli": "1.1.1"
- "@faustwp/core": "1.1.2"
- WordPress version: 6.3

## Captured
<img width="1041" alt="Screen Shot 2023-10-17 at 19 00 12" src="https://github.com/wpengine/faustjs/assets/2823242/86673365-cfa4-4304-9fab-2d14a98390f1">

blakewilson:
Hi @idmahbub, 

Thanks for opening an issue.

Have you verified the query and variables you used in your Faust Template don't return null in the WPGraphQL GraphiQL in the WP Admin? We'll want to rule this out before we can confirm this is a Faust specific issue.


Can you please also paste the query and variables you used in your Faust template? I could test it out locally as well.

arnautg:
This is caused by the new update of the wp-graphql `v1.17.0` downgrading to the `v1.16.0` solves the issue, the problem is that the URI that is passed by parameter is now the full the path `http://localhost:3000/category/news/` when previously was `/category/news` if you run in the GraphQL IDE the following query 
```query GetNodeByUri($uri: String!) {
  nodeByUri(uri: $uri) {
    __typename
  }
}```


Using the full path results into a Post (specifically the latest post of the category)
```
{
  "uri": "/category/noticias/"
}
```

Meanwhile if you run it with this variable results in the category
```
{
  "uri": "/category/noticias/"
}
```

Another temporary fix is to change the `Component.variables` into this:
```Component.variables = ({ uri }) => {
  return {
    uri: new URL(uri).pathname,
    headerLocation: MENUS.PRIMARY_LOCATION,
  };
};```

Not sure which side should fix it

blakewilson:
This could also be related to this comment: https://github.com/wp-graphql/wp-graphql/pull/2341#issuecomment-1128719980

jasonbahl:
@josephfusco and I are working on reproducing right now. 👀

josephfusco:
Hey @idmahbub can you please provide the [Faust.js WordPress Plugin](https://wordpress.org/plugins/faustwp/) version you are using? Thank you!

idmahbub:
[...]

PR #1616, opened 2023-10-18
fix: term_link filter

jasonbahl:
## Description

This fixes the term_link filter callback to not modify the term link urls for GraphQL requests. 

## Related Issue(s):

fixes: #1613

#### BEFORE:

Term uri is absolute in the GraphQL Response

CleanShot 2023-10-18 at 12 51 51

#### AFTER:

The term uri is relative in the GraphQL response

CleanShot 2023-10-18 at 12 53 00

But in the admin UIs the term links are absolute with the alternate Faust front-end domain.

CleanShot 2023-10-18 at 12 53 20

PR #1653, opened 2023-11-14
fix: term_link filter

jasonbahl:
## Description

This fixes the term_link filter callback to not modify the term link urls for GraphQL requests. 

## Related Issue(s):

fixes: #1613

replaces: https://github.com/wpengine/faustjs/pull/1616

#### BEFORE:

Term uri is absolute in the GraphQL Response

CleanShot 2023-10-18 at 12 51 51

#### AFTER:

The term uri is relative in the GraphQL response

CleanShot 2023-10-18 at 12 53 00

But in the admin UIs the term links are absolute with the alternate Faust front-end domain.

CleanShot 2023-10-18 at 12 53 20

### DIFF
--- a/plugins/faustwp/includes/replacement/callbacks.php
+++ b/plugins/faustwp/includes/replacement/callbacks.php
@@ -223,6 +223,11 @@ function term_link( $term_link ) {
 		! is_rewrites_enabled()
 		|| ( function_exists( 'is_graphql_request' ) && is_graphql_request() )
 	) {
+		return equivalent_frontend_url( $term_link );
+	}
+		! is_rewrites_enabled()
+		|| ( function_exists( 'is_graphql_request' ) && is_graphql_request() )
+	) {
 
 	return equivalent_frontend_url( $term_link );
 }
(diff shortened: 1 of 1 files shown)


plugins/faustwp/includes/replacement/callbacks.php
```diff
@@ -219,9 +219,10 @@ function post_link( $link ) {
  * @return string
  */
 function term_link( $term_link ) {
-	if ( ! is_rewrites_enabled() ) {
-		return $term_link;
-	}
+	if (
```
mindctrl, line 222: @jasonbahl there's a syntax error here due to a missing `}`. This should fix it.

```suggestion
	if (
		! is_rewrites_enabled()
		|| ( function_exists( 'is_graphql_request' ) && is_graphql_request() )
	) {
		return equivalent_frontend_url( $term_link );
	}
```

theodesp (approved):
LGTM!

mindctrl (approved):

PR #1659, opened 2023-11-17
Add client utils to app router package

blakewilson:
## Tasks

### DIFF
--- a/examples/next/app-router/app/layout.tsx
+++ b/examples/next/app-router/app/layout.tsx
@@ -1,7 +1,14 @@
 import { gql } from '@apollo/client';
 import { getClient } from '@faustwp/experimental-app-router';
 import Link from 'next/link';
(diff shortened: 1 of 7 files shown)


examples/next/app-router/tsconfig.json
```diff
@@ -13,8 +13,8 @@
     "noEmit": true,
     "incremental": true,
     "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
```
blakewilson, line 17: In order to use package exports in TypeScript, you must set `moduleResolution` to `Node16`, `NodeNext`, or `Bundler`. Since we don't want to enforce strict module resolution with file extensions, etc. on our users, we use `Bundler` since the other two require that strict behavior.

theodesp (approved):

PR #1677, opened 2023-12-11
MERL-1342: removed logging of the telemetry in cli

matthewguywright:
## Description
Removed prompt and log for telemetry. All telemetry scaffolding is still in place but won't fire since they didn't enable telemetry at the cli level.

## Testing
Run the CLI and ensure no messaging for telemetry shows up in the logging.


mindctrl (changes requested):
The lint_packages job failed due to the unused import of `infoLog`. The e2e tests failed too, but seems like a one-off thing.

blakewilson (changes requested):
We have some unused dependencies now after removing telemetry stuff. We'll want to take care of that.

Other than that, looks good @matthewguywright!

blakewilson:
@matthewguywright It looks like we'll need a changeset as well.

blakewilson:
@matthewguywright It looks like `uuid` still exists as a dependency.

matthewguywright:
Mergining to feature/mw-cli-api-v2

PR #1684, opened 2023-12-15
Create process telemetry rest route

blakewilson:
## Tasks

## Description

This PR creates a REST endpoint that accepts a body from the CLI, merges that data with data from WordPress, and then sends it off to GA if the user has Faust telemetry activated on WordPress.

The body looks like:

```
{
    "node_faustwp_core_version": "xxxx",
    "node_faustwp_cli_version": "xxxx",
    "node_faustwp_blocks_version": "xxxx",
    "node_apollo_client_version": "xxxx",
    "node_version": "xxxx",
    "node_next_version": "xxxx",
    "node_is_development": "xxxx",
    "command": "xxxx"
}
```

## Testing

1. Checkout this branch
2. Make a request to `/wp-json/faustwp/v1/process_telemetry` and see the `401` response since you didn't pass a secret key
3. Add the secret key as the header `x-faustwp-secret` and verify you got a `204` response (no content).
4. Modify the `get_telemetry_client_id` function to return a string and notice a `201` is returned (created).
5. Try to break it

### DIFF
--- a/plugins/faustwp/includes/rest/callbacks.php
+++ b/plugins/faustwp/includes/rest/callbacks.php
@@ -110,11 +110,11 @@ function register_rest_routes() {
 		'faustwp/v1',
 		'/process_telemetry',
 		array(
-			'methods' => 'POST',
-			'callback' => __NAMESPACE__ . '\\handle_rest_process_telemetry_callback',
+			'methods'             => 'POST',
+			'callback'            => __NAMESPACE__ . '\\handle_rest_process_telemetry_callback',
 			'permission_callback' => __NAMESPACE__ . '\\rest_process_telemetry_permission_callback',
 		)
-		);
+	);
 
 	/**
 	 * Faust.js packages now use `faustwp/v1/authorize`.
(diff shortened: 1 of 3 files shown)


plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -202,6 +290,19 @@ function rest_telemetry_permission_callback( \WP_REST_Request $request ) {
...
+ *
+ * Authorized if the 'secret_key' settings value and http header 'x-faustwp-secret' match.
+ *
+ * @param \WP_REST_Request $request The current \WP_REST_Request object.
+ *
+ * @return bool True if current user can, false if else.
+ */
+function rest_process_telemetry_permission_callback( \WP_REST_Request $request ) {
+	return rest_authorize_permission_callback( $request );
+}
```
blakewilson, line 304: @TeresaGobble When we save a user's preferences for their telemetry opt-in, we'll also want to generate a unique id for them (something like a UUID) as GA requires this. You can take a look at our previous implementation in the CLI for inspiration on how this can be done.

plugins/faustwp/includes/telemetry/functions.php
```diff
@@ -103,3 +103,18 @@ function get_plugin_version() {
...
+	/**
+	 * @TODO Upon saving the site's telemetry decision, if they accept, we'll
+	 * also need to generate a unique, anonymous client ID for them to be sent 
+	 * with GA requests.
+	 * 
+	 * If a string is returned, telemetry is enabled and a client id has been generated.
+	 * If this function returns null, either telemetry is off, or a client ID is not created.
+	 */
+ return null;
+}
```
blakewilson, line 120: @TeresaGobble When we save a user's preferences for their telemetry opt-in, we'll also want to generate a unique id for them (something like a UUID) as GA requires this. You can take a look at our previous implementation in the CLI for inspiration on how this can be done.

mindctrl: @blakewilson thoughts on using `wp_generate_uuid4()` for this?

blakewilson: @mindctrl I think that's perfect!

theodesp (approved):
LGTM!

PR #1687, opened 2023-12-18
MERL-1343/MERL-1342: API rewrite to WP plugin vs GA

matthewguywright:
## Description
Updated endpoint and refactored to post call to API endpoint for WP plugin vs previous GA4 endpoint.

## Testing
Ensure CLI pings new endpoint for `/wp-json/faustwp/v1/process_telemetry` instead of ga4.


packages/faustwp-cli/src/telemetry/sendTelemetryData.ts
```diff
@@ -1,39 +1,18 @@
...
-      measurement_id: GA_TRACKING_ID,
-    }).toString()}`,
-    {
-      method: 'POST',
-      body: JSON.stringify(body),
-    },
-  );
+export const sendTelemetryData = (payload: TelemetryData) => {
+  return fetch(WP_TELEMETRY_ENDPOINT, {
+    method: 'POST',
```
blakewilson, line 15: This should only be sent with the `FAUST_SECRET_KEY` as a header:

```suggestion
    method: 'POST',
    headers: {
      'x-faustwp-secret': getWpSecret(),
      'Content-Type': 'application/json'
    }
```

blakewilson: Additionally, we should probably only send telemetry data if the WordPress site is over `https` or it's a local site, since we will always be sending this data to WordPress because that is where the telemetry decision is made.

cc @mindctrl

blakewilson (changes requested):

packages/faustwp-cli/src/index.ts
```diff
@@ -102,33 +67,12 @@ import {
...
-        userConfig.get('telemetry.anonymousId') as string,
-      );
-    } catch (err) {
-      // console.log(err);
-      // Fail silently
-    }
+  try {
+    const telemetryData = marshallTelemetryData(arg1);
+    void sendTelemetryData(telemetryData);
+  } catch (err) {
```
blakewilson, line 73: You may need to check here if the secret key exists or not. We do not want to fire off a telemetry event without the secret key in the header.

blakewilson:
@matthewguywright It looks like some tests are failing

.changeset/thick-pugs-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/cli': major
+---
+
+Removed Google Analytics endpoint and replaced with updated plugin API endpoint.
```
mindctrl, line 5: Should we document in the changelog here why this is a breaking change? For example: "We removed <insert the function names we removed here>", so people can easily understand what is breaking without having to dig through PRs, or open support questions, or otherwise fret about what's going on?

.changeset/tough-papayas-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/cli': major
+---
+
+Removed telemetry CLI commands and functionality.
```
mindctrl, line 5: Can we merge these two changesets into one? They're for the same package.

.changeset/thick-pugs-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/cli': major
+---
+
+Removed telemetry CLI commands for `faust telemetry enable/disable/status` for opting in/out and enabling telemetry as well as receiving telemetry status. Telemetry functionality has been moved to the Faust WordPress plugin.
```
mindctrl, line 5:
```suggestion
**BREAKING**: Removed telemetry CLI commands for `faust telemetry enable/disable/status` for managing and viewing telemetry opt-in status. Telemetry functionality has been moved to the Faust WordPress plugin. Sites that previously opted in from CLI will no longer send telemetry data unless someone opts in from the WordPress side.
```

packages/faustwp-cli/src/telemetry/sendTelemetryData.ts
```diff
@@ -1,39 +1,23 @@
...
 
+import { getWpUrl } from '../utils/getWpUrl.js';
+import { getWpSecret } from '../utils/getWpSecret.js';
 import { TelemetryData } from './marshallTelemetryData.js';
-
-const GA_TRACKING_ENDPOINT = 'https://www.google-analytics.com/mp/collect';
-const GA_TRACKING_ID = 'G-KPVSTHK1G4';
-const GA_API_SECRET = '-SLuZb8JTbWkWcT5BD032w';
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+const WP_TELEMETRY_ENDPOINT = `${getWpUrl()!}/wp-json/faustwp/v1/process_telemetry`;
```
blakewilson, line 7: It looks like the requests that are being sent are failing because `getWPUrl()` is returning nothing here. (You can confirm this by commenting out the `console.log` in the `catch` block [here](https://github.com/wpengine/faustjs/pull/1687/files#diff-8c3b33235dbe1bbcfadb06a8c95846376dbe4999ceefe0276219ba239740119fR80).

This is because the `dotenv` config is being instantiated from within the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE#) in the index file, and `WP_TELEMETRY_ENDPOINT` is defined outside of it. You can move `WP_TELEMETRY_ENDPOINT` within the `sendTelemetryData` function and it should work as expected.

packages/faustwp-cli/src/telemetry/sendTelemetryData.ts
```diff
@@ -1,39 +1,23 @@
...
-      measurement_id: GA_TRACKING_ID,
-    }).toString()}`,
-    {
-      method: 'POST',
-      body: JSON.stringify(body),
+export const sendTelemetryData = (payload: TelemetryData) => {
+  return fetch(WP_TELEMETRY_ENDPOINT, {
+    headers: {
+      'x-faustwp-secret': getWpSecret() || '',
+      'Content-Type': 'application/json',
```
blakewilson, line 18: Instead of returning an empty string here, let's throw an error instead of `getWpSecret` is undefined, since we always want this telemetry data to be sent with the secret key

```suggestion
  if(!getWpSecret()) {
    throw new Error('Faust secret key is required');
  }
  
  return fetch(WP_TELEMETRY_ENDPOINT, {
    headers: {
      'x-faustwp-secret': getWpSecret(),
      'Content-Type': 'application/json',
```

blakewilson (changes requested):

blakewilson (approved):
Tested and works as expected 👍  @matthewguywright

PR #1690, opened 2023-12-20
MERL-1339: add WP notice for anonymous telemetry opt-in

TeresaGobble:
## Tasks

## Description

[MERL-1339](https://wpengine.atlassian.net/browse/MERL-1339)

## Testing
- Check out this branch, navigate to the Faust settings page under Settings > Faust.
- You should see a notice asking you to opt in.
- Use the buttons to test the functionality for toggling the setting.
- You should also see a checkbox setting under Features which controls the telemetry setting.
- Play with these various settings and ensure they work as expected. Try to break it.

[MERL-1339]: https://wpengine.atlassian.net/browse/MERL-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -302,3 +314,53 @@ function wpac_authorize_permission_callback( \WP_REST_Request $request ) {
...
+			break;
+	}
+
+	$response = array(
+		'decision' => $decision,
+		'settings' => array(
+			'enabled'  => faustwp_get_setting( 'enable_telemetry' ),
+			'reminder' => faustwp_get_setting( 'telemetry_reminder' ),
+		),
+		'success'  => true,
```
mindctrl, line 363: `faustwp_update_setting()` does not return a value, hence this is currently always returning `true`. We need to better than this. A couple ideas:
- Fetch the value from the database after saving and compare to user-provided decision to determine success.
- Update `faustwp_update_setting()` to return what `update_option()` returns. This would require additional logic, because `update_option()` returns `false` if the values are the same as existing values, and we don't want to perceive this as a "failure" to update.

.changeset/famous-steaks-enjoy.md
blakewilson, line 1: Aptly named changeset @mindctrl

plugins/faustwp/includes/telemetry/assets/js/telemetry-decision-handler.js
blakewilson, line 1: This is some cleannn JS

blakewilson (approved):
Looks great @mindctrl. I tested several different cases and confirmed the correct values were being saved to the db via Adminer. 

The one thing we'll need to do is hook up some anonymous client id field for Google Analytics that gets saved when they either enable telemetry or activate Faust, like the secret key. But that's outside of the scope of this PR

PR #1693, opened 2023-12-21
Generate UUID v4 for telemetry ID

mindctrl:
## Tasks

## Description

Generates and saves a uuid for use with anonymous opt-in telemetry requests.

## Testing

Check out this branch and do one or both of these:
- run `get_telemetry_client_id()` and notice it returns a uuid.
- make a mock telemetry call to the REST route

Note there is a uuid stored in `wp_options.faustwp_settings`.

### DIFF
--- a/plugins/faustwp/includes/settings/callbacks.php
+++ b/plugins/faustwp/includes/settings/callbacks.php
@@ -449,7 +449,7 @@ function display_enable_disable_fields() {
 			<input type="checkbox" id="enable_telemetry" name="faustwp_settings[enable_telemetry]" value="1" <?php checked( $enable_telemetry ); ?> />
 			<?php esc_html_e( 'Enable anonymous telemetry', 'faustwp' ); ?>
 		</label>
-		<input type="hidden" id="telemetry_client_id" name="faustwp_settings[telemetry_client_id]" value="<?php echo get_telemetry_client_id(); ?>" />
+		<input type="hidden" id="telemetry_client_id" name="faustwp_settings[telemetry_client_id]" value="<?php echo esc_attr( get_telemetry_client_id() ); ?>" />
 	</fieldset>
 	<?php
 }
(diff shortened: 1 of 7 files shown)


blakewilson (changes requested):
It looks like if the UUID isn't set, it falls back to `generate_telemetry_client_id`, which is good, but that value is not saved or persisted, so the value will change every request. We'll probably want to save the value if one doesn't exist.

https://github.com/wpengine/faustjs/assets/5946219/0d1b40c3-7c20-422e-ac0a-de69106633b1

mindctrl:
@blakewilson haha yeah I pushed a commit right before you posted! Nice catch. 😅

plugins/faustwp/includes/settings/callbacks.php
```diff
@@ -239,6 +239,14 @@ function sanitize_faustwp_settings( $settings, $option ) {
...
 				break;
 
+			case 'telemetry_client_id':
+				if ( $value ) {
+					$settings[ $name ] = sanitize_text_field( $value );
+				} else {
+					unset( $settings[ $name ] );
+				}
+				break;
+
```
mindctrl, line 249: This is a duplicate of the `enable_telemetry` case. But, it's very important we don't mess that one up, so I kept them separate in case we update one or the other.

blakewilson (commented):
Nice, I've confirmed that worked and the telemetry id is now persisted.

One other small thing I noticed however. This is up to interpretation, but I noticed when telemetry is disabled, then enabled again, the telemetry id is re-generated. Should this be the case?

The telemetry id is meant to anonymously establish a given site, so in theory if someone has telemetry on, disabled it, then re-enables, to us it would look like 2 seperate sites 🤔

mindctrl:
Good catch once again. This is because the uuid was not present in the `$settings` array in the sanitizer, because it wasn't in the `$_POST` data. I've added it as a hidden input so it's always present: 8384b51b2084a57481cbe47eb6d494db43d2ca93

blakewilson (approved):
Awesome! Works great

PR #1720, opened 2024-01-11
Fix anonymous opt-in 

mindctrl:
## Description

Unblocks the `wp_remote_post()` call for anonymous opt-in telemetry.

Fixes a couple other small issues.

## Testing
If you are opted in under Faust > Settings, any of the npm scripts (`dev`, `build`, `generate`, etc.) should result in an event showing up in GA, with associated telemetry data.

### DIFF
--- a/packages/faustwp-cli/src/index.ts
+++ b/packages/faustwp-cli/src/index.ts
@@ -75,10 +75,9 @@ import { marshallTelemetryData, sendTelemetryData } from './telemetry/index.js';
   if (getWpSecret()) {
     try {
       const telemetryData = marshallTelemetryData(arg1);
-      void sendTelemetryData(telemetryData);
+      await sendTelemetryData(telemetryData);
     } catch (err) {
-      // console.log(err);
-      // Fail silently
+      console.log(err);
     }
   }
 
--- a/packages/faustwp-cli/src/index.ts
+++ b/packages/faustwp-cli/src/index.ts
@@ -77,7 +77,10 @@ import { marshallTelemetryData, sendTelemetryData } from './telemetry/index.js';
       const telemetryData = marshallTelemetryData(arg1);
       await sendTelemetryData(telemetryData);
     } catch (err) {
-      console.log(err);
+      debugLog(
+      `Telemetry event failed: `,
+      err,
+    );
     }
   }
 
--- a/packages/faustwp-cli/src/index.ts
+++ b/packages/faustwp-cli/src/index.ts
@@ -77,10 +77,7 @@ import { marshallTelemetryData, sendTelemetryData } from './telemetry/index.js';
       const telemetryData = marshallTelemetryData(arg1);
       await sendTelemetryData(telemetryData);
     } catch (err) {
-      debugLog(
-      `Telemetry event failed: `,
-      err,
-    );
+      debugLog(`Telemetry event failed: `, err);
     }
   }
 
--- a/plugins/faustwp/includes/rest/callbacks.php
+++ b/plugins/faustwp/includes/rest/callbacks.php
@@ -258,7 +258,12 @@ function handle_rest_process_telemetry_callback( \WP_REST_Request $request ) {
 		'multisite'                                    => is_multisite(),
 		'php_version'                                  => PHP_VERSION,
 		'wp_version'                                   => get_wp_version(),
+		'engagement_time_msec'                         => 100,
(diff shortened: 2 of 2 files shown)


packages/faustwp-cli/src/index.ts
```diff
@@ -75,10 +75,9 @@ import { marshallTelemetryData, sendTelemetryData } from './telemetry/index.js';
   if (getWpSecret()) {
     try {
       const telemetryData = marshallTelemetryData(arg1);
-      void sendTelemetryData(telemetryData);
+      await sendTelemetryData(telemetryData);
```
mindctrl, line 78: Without this, the http call wasn't reliably making it to WP before the `process.exit()` call below.

plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -278,18 +283,13 @@ function handle_rest_process_telemetry_callback( \WP_REST_Request $request ) {
...
-	// @codingStandardsIgnoreStart
-	// wp_remote_post($ga_telemetry_url, [
-	// 'body' => $telemetry_body,
-	// ]);
-  // @codingStandardsIgnoreEnd
+	wp_remote_post(
+		$ga_telemetry_url,
+		array(
+			'body'     => wp_json_encode( $telemetry_body ),
+			'blocking' => false,
```
mindctrl, line 290: Added this so WP doesn't wait before continuing. WP says this about the blocking param:

> Whether the calling code requires the result of the request.
If set to false, the request will be sent to the remote server,
and processing returned to the calling code immediately, the caller
will know if the request succeeded or failed, but will not receive
any response from the remote server. Default true.

plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -258,8 +258,13 @@ function handle_rest_process_telemetry_callback( \WP_REST_Request $request ) {
 		'multisite'                                    => is_multisite(),
 		'php_version'                                  => PHP_VERSION,
 		'wp_version'                                   => get_wp_version(),
+		'engagement_time_msec'                         => 100,
+		'session_id'                                   => md5( get_telemetry_client_id() ),
```
mindctrl, line 262: From GA docs:

> In order for user activity to display in reports like [Realtime](https://developers.google.com/analytics/devguides/reporting/data/v1/realtime-basics), engagement_time_msec and session_id must be supplied as part of the params for an event.

Due to delays in updating events in the GA UI, having 'realtime' work was the only reliable way I could determine that "it worked" without waiting an arbitrary/unknown amount of time for the Events page to show.

The way this `session_id` is set here, it would always be the same (and still unique), but maybe we want a way to roll it. From GA docs:

> In Analytics, a session initiates when a user either opens your app in the foreground or views a page or screen and no session is currently active (e.g. their previous session has timed out).

> By default, a session ends (times out) after 30 minutes of user inactivity. There is no limit to how long a session can last.

Thoughts?

packages/faustwp-cli/src/index.ts
```diff
@@ -75,10 +75,9 @@ import { marshallTelemetryData, sendTelemetryData } from './telemetry/index.js';
...
     } catch (err) {
-      // console.log(err);
-      // Fail silently
+      console.log(err);
```
blakewilson, line 80: What do you think about using our `debugLog` here so the error is only present when Faust debug mode is enabled?

```
import { debugLog } from './stdout/index.js';
```

Something like:

```suggestion
      debugLog(
      `Telemetry event failed: `,
      err,
    );
```

plugins/faustwp/includes/rest/callbacks.php
blakewilson: I don't see any downsides in having an indefinite session ID based on the user's unique telemetry ID. We are not tracking bounce rate or other time based things since we are not a traditional website, so from my perspective I don't think it matters if we roll it or not.

mindctrl: Good points. Sounds good to me.

packages/faustwp-cli/src/index.ts
theodesp: Is there a way to use an eslint rule to capture that check?

https://typescript-eslint.io/rules/promise-function-async/

https://github.com/typescript-eslint/typescript-eslint/issues/4553

PR #1724, opened 2024-01-12
Fix 401 token request in WordPressTemplate

### LINKED ISSUE #1722: [Bug] Faust Core: useAuth hook is getting called by default for unauthenticated users triggering extra API calls
# Versions affected

@faustwp-core: 2.0.0

# Description
When navigating to any page you can see extra API calls to the Faust auth token endpoint when the user is not authenticated.

# Steps to reproduce

1. Set experimental toolbar: `false`
2. Visit any page.
3. You should see a call to the `/api/faust/auth/token` endpoint even though the user is not authenticated.

Image

# Technical Details
* There is a call to the useAuth hook be default on the `WordPressTemplate` component here:

[`WordPressTemplate`](https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/components/WordPressTemplate.tsx#L193-L196)

However this happens on every render so it produces additional API calls that can trigger service degradation
We should restore the original functionality which by default it does not trigger this extra auth call.

* When experimental toolbar: `true` then the toolbar also triggers an API call. This could be addressed in another ticket.
* We should ensure authorization only when the page requested is `Preview`

blakewilson:
## Tasks

## Description

This PR skips the `useAuth` call unless the route is a preview route. Currently, `useAuth` is running on every page, which in turn makes a request to the token endpoint to see if a user is authenticated.

## Testing

1. Checkout this branch
2. `npm install` from the root
3. `npm run build` from the root
4. Run the example project `npm run dev -w @faustwp/getting-started-example` from the root
5. Ensure all pages/posts/previews/etc operate as normal while checking to ensure there is not an additional token request made (there will still be a request for the token if `experimentalToolbar` is enabled)
6. Ensure `useAuth` hook operates as normal/expected with and without the new `skip` option.

## Documentation Changes

Closes #1722

### DIFF
--- a/packages/faustwp-core/tests/hooks/useAuth.test.ts
+++ b/packages/faustwp-core/tests/hooks/useAuth.test.ts
@@ -5,6 +5,7 @@
 import { renderHook } from '@testing-library/react-hooks';
 import fetchMock from 'fetch-mock';
 import { useAuth } from '../../src/hooks/useAuth';
+import { setAccessToken } from '../../src/auth';
 
 describe('useAuth hook', () => {
   const envBackup = process.env;
@@ -15,6 +16,10 @@ describe('useAuth hook', () => {
 
   afterAll(() => {
     process.env = envBackup;
+  });
+
+  afterEach(() => {
+    setAccessToken(undefined, undefined);
   });
 
   it('Provides the proper login url with redirect strategy', async () => {
--- a/packages/faustwp-core/tests/hooks/useAuth.test.ts
+++ b/packages/faustwp-core/tests/hooks/useAuth.test.ts
@@ -22,6 +22,43 @@ describe('useAuth hook', () => {
     setAccessToken(undefined, undefined);
   });
 
+  it('skips authentication check when skip is true', async () => {
+    // Set up the fetch mock but expect it not to be called
+    fetchMock.get(`/api/faust/auth/token`, {
+      status: 200,
+      body: JSON.stringify({ accessToken: 'at', refreshToken: 'rt' }),
+    });
+
+    const { result } = renderHook(() => useAuth({ skip: true }));
+
+    // The hook should skip the authentication check
+    expect(result.current.isAuthenticated).toStrictEqual(null);
+    expect(result.current.isReady).toStrictEqual(false);
+
+    // Ensure the token endpoint was not called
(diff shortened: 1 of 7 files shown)


packages/faustwp-core/tests/hooks/useAuth.test.ts
```diff
@@ -17,6 +18,10 @@ describe('useAuth hook', () => {
     process.env = envBackup;
   });
 
+  afterEach(() => {
+    setAccessToken(undefined, undefined);
+  });
+
```
josephfusco, line 24:
```suggestion

  it('skips authentication check when skip is true', async () => {
    // Set up the fetch mock but expect it not to be called
    fetchMock.get(`/api/faust/auth/token`, {
      status: 200,
      body: JSON.stringify({ accessToken: 'at', refreshToken: 'rt' }),
    });

    const { result } = renderHook(() => useAuth({ skip: true }));

    // The hook should skip the authentication check
    expect(result.current.isAuthenticated).toStrictEqual(null);
    expect(result.current.isReady).toStrictEqual(false);

    // Ensure the token endpoint was not called
    expect(fetchMock.called(`/api/faust/auth/token`)).toBe(false);

    fetchMock.restore();
  });

  it('performs authentication check by default (skip not provided)', async () => {
    // Mock a valid response
    fetchMock.get(`/api/faust/auth/token`, {
      status: 200,
      body: JSON.stringify({ accessToken: 'at', refreshToken: 'rt' }),
    });

    const { result, waitForNextUpdate } = renderHook(() => useAuth());

    await waitForNextUpdate();

    // Default behavior should perform authentication check
    expect(result.current.isAuthenticated).toStrictEqual(true);
    expect(result.current.isReady).toStrictEqual(true);

    fetchMock.restore();
  });

```

josephfusco (changes requested):
This looks good to me, only seeing the 401 when experimentalToolbar: true, as expected. 💯 

I added some testing suggestions to cover the new skip option in useAuth

packages/faustwp-core/tests/hooks/useAuth.test.ts
blakewilson: Awesome @josephfusco! I was just thinking I need to write some tests for this, thank you for doing that.

josephfusco (approved):

mindctrl (approved):

Issue #1723, opened 2024-01-12
Improve testing around authorization calls


ChrisWiegman:
This is related to #1722 

We need to more adequately ensure that 401 calls originating from previews or the toolbar can not hurt site performance on a production site without a logged-in user. This ticket is to add appropriate testing to Faust.js to prevent regressions in this area in the future.

blakewilson:
The current issue with the toolbar and auth in general is that currently a request needs to be made to the `/api/faust/auth/token` endpoint to determine if there is a logged in user. 

Traditionally on a page that should only be for authenticated users, this makes sense: we make a request to the token endpoint and if it returns tokens, we know there is a logged in user, if it returns with a `401` response, we know there is no authenticated user and to redirect them to the login page.

With the toolbar, this model is a little difference, since we would like the toolbar to be applied to every page. So, when the toolbar is enabled, a request is made on each page to check if there is an authenticated user. This is not ideal, as we have seen this additional request per page can have unintended consequences.

The challenge here is that we must hit an endpoint to check for an authenticated endpoint. This endpoint though is quite simple. It essentially checks if there is a refresh token in the cookie. This cookie however is `httpOnly` and `Secure`, meaning it can not be accessed client side for XSS reasons. One of the possible solutions I'm playing with in my head is we can set an additional cookie, that is **not** `httpOnly`, that is a boolean value based on **if** there is a refresh token or not. This is beneficial because we'll be able to check this cookie client side, but also not expose any sensitive data since this cookie will only contain a boolean value.

PR #1740, opened 2024-01-17
Only request token endpoint initially, then use a cookie to determine if there is an authenticated user

blakewilson:
## Tasks

## Description

This PR addresses https://github.com/wpengine/faustjs/issues/1723#issuecomment-1896433873 and creates a new cookie that determines if there is a refresh token saved in the cookie, without exposing sensitive data and is accessible in the browser/client.

This allows us to make one initial request to the token endpoint, and then thereafter reference the cookie to determine if there is an authenticated user for stuff like the toolbar. This essentially turns a request for the token on every single page to only once for non-authenticated users.

## Testing

1. From the `canary` branch, start the example project: `npm run dev -w @faustwp/getting-started-example`
2. Open the network tools, and notice every page you visit has a 401 response for the token endpoint.
3. Now, checkout this branch
4. Run `npm install` at the monorepo root
5. Run `npm run build` at the monorepo root
6. Run the example project `npm run dev -w @faustwp/getting-started-example`
7. With the network tools open, visit the home page http://localhost:3000. Notice the 401 response from the token endpoint. Refresh and see that no additional requests are made to the token endpoint as an unauthenticated user.
8. Try and break it with the other testing things below

### Other testing things to ensure functionality still works as expected

- [ ] Ensure refresh tokens are still set properly on login/logout
- [ ] Ensure backwards compatibility still works. For example, if there is a cookie for just the refresh token, the reference token should still get set properly


blakewilson:
Closes #1723

matthewguywright (approved):

mindctrl (approved):

PR #1788, opened 2024-02-06
Update @apollo/experimental-nextjs-app-support to v0.8.0-OLD

theodesp:
## Tasks

## Description

Updates @apollo/experimental-nextjs-app-support to v0.8.0 in both the package and the example projects

## Testing

1. Cleanup all your node_modules from the `examples/next/app-router`.
2. Run `npm i` and `npm run build`
3. Start the app-router example project. It should works as. expected.

### DIFF
--- a/packages/experimental-app-router/package.json
+++ b/packages/experimental-app-router/package.json
@@ -48,7 +48,7 @@
   },
   "peerDependencies": {
     "@apollo/client": ">=3.8.0",
-    "@apollo/experimental-nextjs-app-support": ">=0.5.0",
+    "@apollo/experimental-nextjs-app-support": ">=0.8.0",
     "@faustwp/cli": ">=1.1.3",
     "@faustwp/core": ">=1.1.2",
     "next": ">=14.0.0",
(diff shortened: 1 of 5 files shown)


packages/experimental-app-router/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "peerDependencies": {
     "@apollo/client": ">=3.8.0",
-    "@apollo/experimental-nextjs-app-support": ">=0.5.0",
```
blakewilson, line 51: I'm not sure if we should update the peer dependency's version here, that would be a breaking change.

https://github.com/semver/semver/issues/502#issuecomment-466501843

By upgrading the example project and the dev dependency we are still avoiding the dependabot vulnerability but giving users the choice to use 0.5.0 through 0.8.0.

theodesp: Its a v1.0 change so the risk is justified. Also we cannot allow the user to have a vulnerable version anyway

blakewilson: Agreed. We should still increment `experimental-app-router` with a `major` bump though.

TeresaGobble:
@theodesp  I tried updating to the fix in my own PR to pass the `Audit Dependencies` check, but it then causes a cascade of failing checks elsewhere, similar to what you have here. I'm not able to pinpoint why- just wanted to confirm that I'm seeing failures after the update attempt as well.

Issue #1761, opened 2024-01-25
Post Previews Page redirects to login page on App Router/Faust.js package 


Fran-A-Dev:
## Environment
- `@faustwp/core` version: "^1.2.0"
- `@faustwp/cli` version: "^1.2.0"
- `@faustwp/blocks` version: 
- `"@faustwp/experimental-app-router": "^0.2.0"`

- WordPress version: 6.3.2

## Description

The issue that I am experiencing is with the Post Previews feature of the experimental App Router package for Faust.js. Upon creating and editing a post in the WP admin and clicking 'preview' to see the edited post before publishing, it redirects me to the login page on the app, and when I log in successfully, it does not show me the post preview page that I am expecting. It goes back to the home page.

I have checked all the necessary settings to ensure there is an auth cookie in the browser, the Faust secret key, and the frontend URL are in place, etc. And it still behaves this way. Here are some screenshots to show you the flow of the bug I am experiencing.

Press the post preview button in WP Admin:
<img width="1728" alt="Screenshot 2024-01-25 at 10 45 55 AM" src="https://github.com/wpengine/faustjs/assets/73023614/72567072-8171-470c-8132-6ed46c8b971a">

Redirects to the login page:
<img width="873" alt="Screenshot 2024-01-25 at 10 46 22 AM" src="https://github.com/wpengine/faustjs/assets/73023614/feda39cb-a896-4890-849a-6954b592b442">

Upon logging in, it redirects back to the home page instead of showing the expected post preview page:
<img width="1728" alt="Screenshot 2024-01-25 at 10 47 01 AM" src="https://github.com/wpengine/faustjs/assets/73023614/6c50f1c0-632c-403f-af92-23f7d423280d">

## Steps to Reproduce

1. Download the Faust.js experimental App Router Package.
2. Set up all necessary environment variables and plugins in WP Admin.
3. Grab the Post Preview code from the canary branch for the experimental App Router and place it in your local codebase: [Faust.js Example Page](https://github.com/wpengine/faustjs/blob/canary/examples/next/app-router/app/%5Bslug%5D/page.tsx)
4. Once configured, go to WP admin and try to add or edit a post and hit the post preview button.

## Expected vs. Actual Behavior

Expected behavior is for the Post Previews feature of WordPress to work in a headless setup with the App Router and Faust.js.

The actual behavior is being redirected to the login page after attempting to use the Post Previews feature, having to log in, and not seeing the post previews page.

## Reproducible Demo

[...]

PR #1799, opened 2024-02-09
1761: Fix preview issue due to samesite status

matthewguywright:
## Description
Update SameSite to use 'lax' instead of 'strict' to allow WordPress urls in Faust routing for previews.

## Testing
[https://github.com/wpengine/faustjs/issues/1761](https://github.com/wpengine/faustjs/issues/1761)
The steps to reproduce are in the linked ticket. But instead of failing it should properly show the post the was previewed.


.changeset/healthy-mayflies-end.md
```diff
@@ -0,0 +1,5 @@
+---
+'@faustwp/experimental-app-router': minor
```
theodesp, line 2: I don't think this is breaking since it fixes a security recommendation which does not change the usage or the way that users should use the package.

matthewguywright: @blakewilson Thoughts?

.changeset/healthy-mayflies-end.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Since app router requests are made server side and the preview post request originates from WordPress, the cookie isn’t reachable. Changing the cookie to SameSite=Lax resolves the problem.
```
blakewilson, line 5: Let's be clearer on what was changed for the end user:

```suggestion
Changed the `sameSite` property on the refresh token cookie from `strict` to `lax` for requests originating from WordPress.
```

.changeset/healthy-mayflies-end.md
blakewilson: @matthewguywright @theodesp  I think that makes sense. The refresh token cookie we use is not part of our "public api" and so modifying it should not come as a surprise to anyone.

blakewilson (changes requested):
@matthewguywright Have you tested what happens if a site with a user who is already logged in (i.e. has a cookie set with `sameSite=strict`) then tries to get a new refresh token and that cookie tries to be re-written to `sameSite=lax`?

theodesp:
> @matthewguywright Have you tested what happens if a site with a user who is already logged in (i.e. has a cookie set with `sameSite=strict`) then tries to get a new refresh token and that cookie tries to be re-written to `sameSite=lax`?

The cookie in that case will still be `sameSite=strict` since there is no change to the cookie if the user is already authenticated. I think that may create issues when opening a new tab for previews so it's important for the user to clear the cookies.

blakewilson:
> > @matthewguywright Have you tested what happens if a site with a user who is already logged in (i.e. has a cookie set with `sameSite=strict`) then tries to get a new refresh token and that cookie tries to be re-written to `sameSite=lax`?
> 
> The cookie in that case will still be `sameSite=strict` since there is no change to the cookie if the user is already authenticated. I think that may create issues when opening a new tab for previews so it's important for the user to clear the cookies.

@theodesp @matthewguywright That's what it looks like. After some quick testing it seems that a cookie can not be updated from `SameSite=Strict` to `SameSite=Lax` so for this feature to take effect, they will have to log out, and log back in. Which I don't think is a problem.

.changeset/healthy-mayflies-end.md
blakewilson: @matthewguywright Since we agreed it's not a breaking change, let's update the changeset type to `patch`

blakewilson (approved):

PR #1806, opened 2024-02-19
16: update lowest version of node to 18 since 16 is deprecating

matthewguywright:
## Description
Update dependency on node 16+ to 18+ since 16 is being deprecated.

## Testing
1. Ensure you have switched to the proper branch and pulled latest
2. Run `npm install` to install node modules based on node 18+
3. Test out the example projects to ensure there are no errors or failures in the build and display
4. Note: this does not include internal/legacy project


.github/workflows/e2e-next-example.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       PR_BRANCH: ${{github.event.pull_request.head.ref}}
     strategy:
       matrix:
-        node: ['16', '18', '20']
+        node: '18', '20']
```
mindctrl, line 17: It looks like all these need the open `[`

```suggestion
        node: ['18', '20']
```

examples/next/faustwp-getting-started/package.json
```diff
@@ -23,7 +23,7 @@
     "sass": "^1.54.9"
   },
   "engines": {
-    "node": ">=16",
+    "node": ">=18",
```
mindctrl, line 26: Is there any downside to bumping to 20? It's the most recent LTS version.

matthewguywright: I wouldn't just because 18 is still a relevant and non deprecated version.

theodesp (approved):

PR #1809, opened 2024-02-21
Feature: Error when NEXT_PUBLIC_WORDPRESS_URL same as headless site URL

### LINKED ISSUE #1750: Faust should warn me if I set my NEXT_PUBLIC_WORDPRESS_URL as my headless URL
As a Faust user, I would like to be warned if I accidentally set the `NEXT_PUBLIC_WORDPRESS_URL` as the URL of my headless site.

We have seen this happen times here and there, and having some logic to warn the user of this behavior would make it easier for them to debug/troubleshoot.

We could make a request to the GraphQL endpoint and if it doesn't return with a proper endpoint, and has an `x-using: faust` header, we know they pointed the URL to the headless site.

## Acceptance Criteria

- [ ] Faust CLI warns the user/quits early if the headless url is specified for `NEXT_PUBLIC_WORDPRESS_URL` instead of their WordPress instance.

theodesp:
## Tasks

## Description

This pull request adds a check in order to alert Faust users if they accidentally set the NEXT_PUBLIC_WORDPRESS_URL to the URL of their headless site. The code now makes a request to the FaustWP endpoint and checks if the response indicates that the URL is pointing to a headless site.

## Related Issue(s):

Closes #1750 

## Testing
To test these changes, follow these steps:

1. In the Faust Plugin Settings, set the Headless URL to match the NEXT_PUBLIC_WORDPRESS_URL.
2. Try to run the dev or build commands in your headless site.
Observe that it should error with the following message:
```
Validation Failed: Ensure your NEXT_PUBLIC_WORDPRESS_URL does not match with Headless URL in the Faust WordPress plugin settings
```

### Testing Instructions (for Older Plugin Versions)
To test these changes on older versions of the Faust Plugin, follow these steps:

1) In the Faust Plugin Settings, set the Headless URL to match the NEXT_PUBLIC_WORDPRESS_URL.
2) Try to run the development or build commands in your headless site.
3) Observe the behavior when the plugin version is outdated:
4) The function should continue silently, logging a message indicating the missing route and prompting the user to update their FaustWP plugin to the latest version.

```
warn - Route not found: Please update your FaustWP plugin to the latest version.
```

## Screenshots

Screenshot 2024-02-21 at 13 11 29

Screenshot 2024-02-21 at 13 12 00

### DIFF
--- a/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
+++ b/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
@@ -1,0 +1,34 @@
+import { getWpSecret, getWpUrl } from '../utils/index.js';
+import { errorLog } from '../stdout/index.js';
+
+/**
+ * Validates the NEXT_PUBLIC_WORDPRESS_URL environment variable by sending a POST request to the Faust Plugin API.
+ * If the URL matches the Faust Plugin Headless URL, the validation fails, and an error is logged.
+ */
+export async function validateNextWordPressUrl(): Promise<void> {
+  const apiUrl = `${getWpUrl()}/wp-json/faustwp/v1/validate_public_wordpress_url`;
+  const headers = {
+    'Content-Type': 'application/json',
+    'x-faustwp-secret': getWpSecret() || '',
+  };
+
+  const postData = {
+    public_wordpress_url: getWpUrl(),
+  };
+  try {
+    const response = await fetch(apiUrl, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify(postData),
+    });
+
+    if (!response.ok) {
+      errorLog(
+        'Validation Failed: Ensure your NEXT_PUBLIC_WORDPRESS_URL does not match with Headless URL in the Faust WordPress plugin settings',
+      );
+      process.exit(1);
+    }
+  } catch (error) {
+    console.log('error', error);
+  }
+}
--- a/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
+++ b/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
@@ -1,5 +1,5 @@
 import { getWpSecret, getWpUrl } from '../utils/index.js';
-import { errorLog } from '../stdout/index.js';
+import { errorLog, warnLog } from '../stdout/index.js';
 
 /**
  * Validates the NEXT_PUBLIC_WORDPRESS_URL environment variable by sending a POST request to the Faust Plugin API.
@@ -23,10 +23,17 @@ export async function validateNextWordPressUrl(): Promise<void> {
     });
 
     if (!response.ok) {
-      errorLog(
-        'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
-      );
-      process.exit(1);
+      if (response.status === 404) {
+        // Handle the case when the route does not exist
+        warnLog(
+          'Route not found: Please update your FaustWP plugin to the latest version.',
+        );
+      } else {
+        errorLog(
+          'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
+        );
+        process.exit(1);
+      }
     }
   } catch (error) {
     console.log('error', error);
--- a/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
+++ b/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
@@ -30,7 +30,7 @@ export async function validateNextWordPressUrl(): Promise<void> {
         );
       } else {
         errorLog(
-          'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
+          'Validation Failed: Your Faust front-end site URL value is misconfigured. It should NOT match the `NEXT_PUBLIC_WORDPRESS_URL.`',
         );
         process.exit(1);
       }
--- a/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
+++ b/packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
@@ -24,7 +24,7 @@ export async function validateNextWordPressUrl(): Promise<void> {
 
     if (!response.ok) {
       errorLog(
-        'Validation Failed: Ensure your NEXT_PUBLIC_WORDPRESS_URL does not match with Headless URL in the Faust WordPress plugin settings',
+        'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
       );
       process.exit(1);
     }
--- a/plugins/faustwp/includes/rest/callbacks.php
+++ b/plugins/faustwp/includes/rest/callbacks.php
@@ -565,11 +565,11 @@ function handle_rest_validate_public_wordpress_url_callback( \WP_REST_Request $r
 			$response = new \WP_REST_Response( 'OK', 200 );
 		} else {
 			// Return 400 Bad Request if the URLs match.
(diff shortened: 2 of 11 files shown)


packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
```diff
@@ -0,0 +1,34 @@
...
+  try {
+    const response = await fetch(apiUrl, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify(postData),
+    });
+
+    if (!response.ok) {
+      errorLog(
+        'Validation Failed: Ensure your NEXT_PUBLIC_WORDPRESS_URL does not match with Headless URL in the Faust WordPress plugin settings',
```
mindctrl, line 27: Thoughts on this wording?

```suggestion
        'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
```

mindctrl (changes requested):
The lint plugin job is flagging a number of phpcs issues.

plugins/faustwp/includes/utilities/functions.php
```diff
@@ -43,3 +43,23 @@ function plugin_version() {
...
 }
+
+/**
+ * Checks if two domain strings represent the same domain.
+ *
+ * @param string $domain1 The first domain string.
+ * @param string $domain2 The second domain string.
+ * @return bool True if the domains match, false otherwise.
+ */
+function domains_match( $domain1, $domain2 ) {
```
mindctrl, line 54: Can you write some tests demonstrating the desired behavior for this function?

mindctrl (changes requested):

TeresaGobble:
Hi @theodesp ! Nice job on this, it's going to be so helpful to folks.

I tested this branch by matching the wp-admin URL to the .env.local URL:
image
image

This is the error I received:
image

Everything seems to be working as expected with that error log! I am a little confused on the warning log's behavior, however. I changed my NEXT_PUBLIC_WORDPRESS_URL to another URL to test changes and noted that the warning prompts an update to Faust's plugin:
image

image

I think it should be rare for folks to have a misspelling or an entirely incorrect NEXT_PUBLIC_WORDPRESS_URL, but if they do, this warning might be confusing.

packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (!response.ok) {
+      if (response.status === 404) {
+        // Handle the case when the route does not exist
+        warnLog(
+          'Route not found: Please update your FaustWP plugin to the latest version.',
+        );
+      } else {
+        errorLog(
+          'Validation Failed: Your NEXT_PUBLIC_WORDPRESS_URL value is misconfigured. It should match your WordPress site URL and not your Faust front-end site URL.',
```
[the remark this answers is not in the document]
theodesp: You are reaching this because of this:

NEXT_PUBLIC_WORDPRESS_URL points to an invalid route, and `http://localhost:3000/index.php?graphql` is not a valid graphQL endpoint.

theodesp:
> I think it should be rare for folks to have a misspelling or an entirely incorrect NEXT_PUBLIC_WORDPRESS_URL, but if they do, this warning might be confusing.

Actually, if you provide an invalid `NEXT_PUBLIC_WORDPRESS_URL` then it can't even reach the graphQL endpoint since the NEXT_PUBLIC_WORDPRESS_URL is invalid. Chicken and Egg problem.

Maybe if I can update the Validation Error message to say something about the Headless Site URL:

```
Validation Failed: Your Faust front-end site URL value is misconfigured. It should not match the `NEXT_PUBLIC_WORDPRESS_URL`
```

packages/faustwp-cli/src/healthCheck/validateNextWordPressUrl.ts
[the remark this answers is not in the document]
mindctrl: @theodesp yah that's because I was testing the issue as explained:

> This pull request adds a check in order to alert Faust users if they accidentally set the NEXT_PUBLIC_WORDPRESS_URL to the URL of their headless site.

But it seems like if that's the case this logic won't be reached due to other logic, unless the front-end URL happens to also have a graphql endpoint?

mindctrl:
>  I changed my NEXT_PUBLIC_WORDPRESS_URL to another URL to test changes and noted that the warning prompts an update to Faust's plugin
> ...
> I think it should be rare for folks to have a misspelling or an entirely incorrect NEXT_PUBLIC_WORDPRESS_URL, but if they do, this warning might be confusing.

I'm guessing the other site doesn't have this branch checked out for the WP plugin? If so, that would be expected behavior since the `/validate_public_wordpress_url` does not exist on the older versions of Faust WP.

blakewilson:
> > I changed my NEXT_PUBLIC_WORDPRESS_URL to another URL to test changes and noted that the warning prompts an update to Faust's plugin
> > ...
> > I think it should be rare for folks to have a misspelling or an entirely incorrect NEXT_PUBLIC_WORDPRESS_URL, but if they do, this warning might be confusing.
> 
> I'm guessing the other site doesn't have this branch checked out for the WP plugin? If so, that would be expected behavior since the `/validate_public_wordpress_url` does not exist on the older versions of Faust WP.

This is a good point. This could be a breaking change if users are not using the latest plugin version.

Is there a way we could do this check without a new Rest endpoint?

For example, making a request to ${process.env.NEXT_PUBLIC_WORDPRESS_URL}/graphql` and check the response to see if it returns a valid GraphQL query, and if it does, suggest that they might be using their frontend URL instead of their WordPress URL.

theodesp:
> Validation Failed: Your Faust front-end site URL value is misconfigured. It should not match the `NEXT_PUBLIC_WORDPRESS_URL`

I don't think this is reasonable enough since the user might setup a graphql endpoint on `${headlessURL}/graphql` on their own, so it will be a false flag.

Its also very confusing since if the `process.env.NEXT_PUBLIC_WORDPRESS_URL` points to an actual WP instance, then a request to `${process.env.NEXT_PUBLIC_WORDPRESS_URL}/graphql` will always succeed!

> This could be a breaking change if users are not using the latest plugin version.

I do check if the response code is 404 which means that the user hasn't updated their plugin to handle the new endpoint so there shouldn't be a breaking change.

blakewilson:
> > This could be a breaking change if users are not using the latest plugin version.
> 
> I do check if the response code is 404 which means that the user hasn't updated their plugin to handle the new endpoint so there shouldn't be a breaking change.

Nice 👍 I missed this at first

plugins/faustwp/includes/rest/callbacks.php
```diff
@@ -523,3 +534,43 @@ function handle_rest_validate_secret_key_callback( \WP_REST_Request $request ) {
...
+		if ( ! domains_match( $public_wordpress_url, $frontend_uri ) ) {
+			// Return 200 OK if the URLs do not match.
+			$response = new \WP_REST_Response( 'OK', 200 );
+		} else {
+			// Return 400 Bad Request if the URLs match.
+			$response = new \WP_REST_Response( 'Bad Request: Provided URL matches the frontend URI setting.', 400 );
+		}
+	} else {
+		// Return 400 Bad Request if the public_wordpress_url parameter is missing.
+		$response = new \WP_REST_Response( 'Bad Request: public_wordpress_url parameter is missing.', 400 );
```
mindctrl, line 572: If returning error messages here, why not use them in the JS code instead of hardcoding the error strings there too?

theodesp: Removed those messages.

mindctrl (changes requested):
There's a bug in the logic here. When your `NEXT_PUBLIC_WORDPRESS_URL` is valid but you don't have a `FAUST_SECRET_KEY` value, you get the following error:

```
warn - Could not find FAUST_SECRET_KEY environment variable.
warn - Some functionality may be limited.
error - Validation Failed: Your Faust front-end site URL value is misconfigured. It should NOT match the `NEXT_PUBLIC_WORDPRESS_URL.`
```

In this case, my WP URL is correct and not pointing to my front-end site URL.

theodesp:
> FAUST_SECRET_KEY

FAUST_SECRET_KEY is required to perform the validation. I will skip the test if this is not provided since it will fail the previews anyway.

mindctrl (approved):

Issue #1903, opened 2024-06-05
Preview detection via query string is too greedy and catches non WP previews


ccorda:
### Description

We're implementing a one trust cookie manager that has preview links that look like this:
`?otreset=true&otpreview=true&otgeo=`

Faust is incorrectly assuming this is a WP preview link, and redirecting logged out users to the WP admin.

This is the code where this is happening:
https://github.com/wpengine/faustjs/blob/canary/packages/faustwp-core/src/components/WordPressTemplate.tsx#L207

### Steps to reproduce

Add `?otpreview=true` to any url, e.g. `https://faustjs.org/?otpreview=true`

### @faustwp/core Version

1.2.1

### @faustwp/cli Version

1.2.1

### FaustWP Plugin Version

1.2.1

### WordPress Version

5.3

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

theodesp:
@ccorda the latest @faustwp/core@3.0.3 version contains a fix for this issue. I'm closing this ticket but feel free to comment back if you still any problems.

PR #1911, opened 2024-06-17
Fix: Ensure Accurate Detection of WordPress Previews via URL Query Parameters

theodesp:
## Tasks

## Description

This PR addresses a bug in the WordPressTemplate component where the detection of WordPress preview URLs was overly broad, potentially catching non-WordPress preview URLs. The fix introduces a more precise method for determining if the URL is a WordPress preview.

## Related Issue(s):

https://github.com/wpengine/faustjs/issues/1903

## Problem
The previous implementation checked if the URL query string contained preview=true using a simple string includes method. This approach could incorrectly identify non-WordPress previews as WordPress previews if they happened to have a similar query parameter structure.

## Solution
* Introduced a utility function isWordPressPreview to accurately check for the presence of preview=true in the URL query parameters.
* Updated the WordPressTemplate component to use this utility function for determining if the URL is a WordPress preview.
* Refactored the related unit tests to reflect these changes and ensure comprehensive test coverage.

## Testing

Faust should not be redirecting logged out users (visitors) to the WP admin  if the page contains a search query that contains the string `preview=`.

For example:

http://localhost:3000/posts/add-table?otpreview=true

### DIFF
--- a/packages/faustwp-core/tests/components/WordPressTemplate.test.tsx
+++ b/packages/faustwp-core/tests/components/WordPressTemplate.test.tsx
@@ -27,56 +27,6 @@ describe('<WordPressTemplate />', () => {
     );
   });
 
-  test('Properly determines whether or not the given URL is a preview or not', async () => {
-    const getConfigSpy = jest.spyOn(getConfig, 'getConfig').mockReturnValue({
-      templates: {},
-    });
-
-    const useAuthSpy = jest.spyOn(useAuth, 'useAuth').mockReturnValue({
-      isAuthenticated: false,
-      isReady: true,
-      loginUrl: null,
-    });
-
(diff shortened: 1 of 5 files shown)


packages/faustwp-core/tests/components/WordPressTemplate.test.tsx
```diff
@@ -27,56 +27,6 @@ describe('<WordPressTemplate />', () => {
     );
   });
 
-  test('Properly determines whether or not the given URL is a preview or not', async () => {
```
theodesp, line 30: Problematic unit test with lots of testing anti-patterns. I removed this since we already test the logic of isWordPressPreview

ChrisWiegman (approved):

Issue #1885, opened 2024-05-10
Sitemaps.xml breaks upon updates made in RankMath


ariel-integrity:
### Description

I'm using the suggested solution in the faust.js docs to create a sitemap for my headless site.

The only issue I've had is that when I make sitemap updates in the RankMath settings, it breaks my CPT sitemaps. 

However, if I turn the faust.js plugin off, clear cache, and turn it back on, they work. But in the meantime, I get errors on pages due to faust.js being turned off. Need a solution where I can skip that step.

### Steps to reproduce

Set up a sitemaps.xml doc in a faust.js project. 

Install RankMath and turn on the sitemap generator. Update the settings to include/not include a CPT.

See that those sitemaps, and maybe the posts/pages, break. Turn off faust.js, clear cache, and turn it back on. They work!

### Additional context

Site is up in an Atlas environment.

### @faustwp/core Version

1.2.0

### @faustwp/cli Version

1.2.1

### FaustWP Plugin Version

1.2.3

### WordPress Version

6.5.2

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

ChrisWiegman:
Hello @ariel-integrity! 

My apologies for the late reply here. I've attempted to reproduce the issue with the free version of Rankmath and have been unable to do so.

A few things you might want to check:

1. Verify the front-end URL env variable is set correctly in `sitemap.xml.js`. 
2. Ensure your permalinks are not set to `Plain`.

Can you please verify for me that your sitemap is still broken with the above settings?

ariel-integrity:
@ChrisWiegman Sorry for the late reply. I am facing this issue once again, but have very much narrowed it down.

My sitemaps only seem to break when I add a new post or custom post type post. Other than that, they work correctly and as expected based on the faust.js sitemaps documentation.

I have an example up and running on our dev DB - let me know if there is a secure way to send it to you.

In regards to front-end URL and permalinks:
My front-end URL is correctly set, and I am able to verify this, because the vast majority of sitemaps and links are all correct the vast majority of the time.
My permalinks are set to custom.

josephfusco:
Hey @ariel-integrity,

Could you please clarify what you mean by “the sitemap breaks”? For example, is it producing invalid XML, are you getting a 404 error, or something else? The more details you can provide, the better we’ll be able to assist you.

----
Just a heads-up, ChrisWiegman is no longer working on this project.

ariel-integrity:
[...]

PR #1936, opened 2024-08-20
Introduce configurable "sitemap index path" for enhanced plugin compatibility

josephfusco:
## Description

This PR introduces a new feature allowing the configuration of a custom sitemap index path via the `sitemapIndexPath` option in the getSitemapProps function. This is particularly useful for ensuring compatibility with WordPress plugins like RankMath that modify the default sitemap path.

To configure a custom sitemap index path, update your getServerSideProps function in your Next.js page as follows:

```js
import { getSitemapProps } from '@faustwp/core';

export default function Sitemap() {}

export function getServerSideProps(ctx) {
  return getSitemapProps(ctx, {
    sitemapIndexPath: '/sitemap_index.xml', // Specify the custom sitemap index path
    frontendUrl: process.env.NEXT_PUBLIC_SITE_URL,
  });
}
```

This is particularly useful when using WordPress plugins like RankMath that change the default sitemap path from `/wp-sitemap.xml` to another path, such as `/sitemap_index.xml`.

## Related Issue(s):

- #1885

## Testing

1. Enable RankMath and ensure it changes the sitemap index path to `/sitemap_index.xml`.
2. Update getServerSideProps with sitemapIndexPath: `/sitemap_index.xml`.
3. Deploy the application.
4. Visit /sitemap.xml and verify the sitemap loads correctly.
5. Confirm all expected URLs are present and no 404 errors occur.

### DIFF
--- a/.changeset/long-humans-type.md
+++ b/.changeset/long-humans-type.md
@@ -1,0 +1,19 @@
+---
+'@faustwp/core': minor
+---
+
+- Added support for configuring a custom sitemap index path via the `sitemapIndexPath` option in `getSitemapProps`, enhancing compatibility with plugins like RankMath that modify the default sitemap path.
+
+  ```javascript
+  import { getSitemapProps } from '@faustwp/core';
+
+  export default function Sitemap() {}
+
+  export function getServerSideProps(ctx) {
+    return getSitemapProps(ctx, {
+      sitemapIndexPath: '/sitemap_index.xml', // RankMath changes the default sitemap path to this
+      frontendUrl: process.env.NEXT_PUBLIC_SITE_URL,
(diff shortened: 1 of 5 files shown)


theodesp (approved):
LGTM!

cedricrabarijohn:
@josephfusco I suggest a full customization of the url too, not only the sitemapIndexPath. It's usefull when your graphql endpoint is not by default accessible via https://wp-url.com but via https://wp-url.com/wp for example. And there will be no workaround for the Next project if the sitemap of the wordpress app is inside https://url.com/sitemap_index.xml and not accessible in https://url.com/wp/sitemap_index.xml. So instead of only allowing the app to use the NEXT_PUBLIC_WORDPRESS_URL env variable , maybe having an option to use a completely different url and override the default one would be usefull.

PR #1939, opened 2024-08-20
fix: ssr not always working

### LINKED ISSUE #1813: SSR only works when template query is present
### Description

When a template component doesn't have a `query` property, the SSR for that page simply doesn't work.

It looks like a bug to me because it doesn't make sense – if a page doesn't require remote data, there's no reason not to generate it statically.

In my reproduction example, I added a query to fetch some unused piece of data, just to show that the sole presence of a query triggers the SSR.

### Steps to reproduce

1. Clone my [reproduction repository](https://github.com/rrmesquita/wpengine-faustjs-ssr-issue)
2. Install dependencies, run the build then `npm run start` to start the server
3. In the browser, navigate to [/sample-page](http://localhost:3000/sample-page)
4. Right-click and "View Page Source"
5. Search for "testing" and assert **it is not** there
6. Stop the server, go to `wp-templates/page.js`, uncomment the template query and save the file
7. Run the build and start the server again
8. Update the View Source page
9. Search for "testing" and assert **it is** there.

### Additional context

Page Source screenshot:

Screenshot 2024-02-21 at 22 02 59

### @faustwp/core Version

2.1.2

### @faustwp/cli Version

2.0.0

### FaustWP Plugin Version

Using hosted example

### WordPress Version

Using hosted example

### Additional environment details

Using hosted example

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

jasonbahl:
## Description

This fixes a bug where the `WordPressTemplate` component would return `null` on the server when `isPreview` was null.

## Related Issue(s):

fixes #1813 

## Testing

I tested following the steps outlined in #1813 on the reproduction mentioned in the issue. 

I also was able to reproduce the issue on acf.wpgraphql.com (codebase here: https://github.com/wp-graphql/acf.wpgraphql.com) and was able to confirm that the issue is resolved for acf.wpgraphql.com with this fix in place. 

### Reproduction Site from #1813 - Before

With JavaScript disabled, the page is blank

CleanShot 2024-08-20 at 15 51 32

### Reproduction Site from #1813 - After

Content is rendered even with JavaScript disabled

CleanShot 2024-08-20 at 15 54 20

### acf.wpgraphql.com - Before

With JavaScript disabled, the page is blank

CleanShot 2024-08-20 at 15 50 31

### acf.wpgraphql.com  - After

Content is rendered even with JavaScript disabled

CleanShot 2024-08-20 at 15 57 02

### DIFF
--- a/.changeset/tiny-years-grin.md
+++ b/.changeset/tiny-years-grin.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/core': patch
+---
+
+Fixes a bug where the WordPressTemplate component would return `null` on the server when `isPreview` was null.
--- a/.changeset/wet-camels-lick.md
+++ b/.changeset/wet-camels-lick.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/cli': patch
+---
+
+Added "glob" as a dependency
--- a/package-lock.json
+++ b/package-lock.json
@@ -3453,7 +3453,6 @@
     },
     "node_modules/@apollo/client": {
       "version": "3.8.6",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@graphql-typed-document-node/core": "^3.1.1",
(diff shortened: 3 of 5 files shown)


josephfusco (approved):
I was able to see this working on both the homepage and the upgrade guide with JavaScript turned off.

CleanShot 2024-08-21 at 16 33 19@2x

CleanShot 2024-08-21 at 16 33 57@2x

PR #1985, opened 2024-11-13
Adding a filter for optional change of public redirect status codes

infn8:
Adding support for users to change the redirect status code from a `302` to another code of their choice, likely `301`.

## Tasks

## Description

Added new filter `faustwp_public_redirect_status_code`, allowing WordPress plugins and themes to choose the [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) to use when generating redirects when the [enable public route redirects](https://faustjs.org/docs/faustwp/settings#enabling-public-route-redirects) setting is active.

## Related Issue(s):

None

## Testing

I added this locally and ensured that the redirects were still `302` status codes.  I then added this in my theme's `functions.php` file to test it

```php
add_filter('faustwp_public_redirect_status_code', function ($status_code) {
	return 301;
}, 10, 1);
```

Then I tried calling the exact same URL which resulted in a `301` status code with _no change_ to other headers including the `x-redirect-by: WP Engine Headless plugin` header

## Screenshots

none

## Documentation Changes

There is nowhere in the repo docs that cover this functionality but there _is_ the following on the live site:

https://faustjs.org/reference/faust-wordpress-plugin-filters

So I am writing here markdown that will get you close, if not all the way to an update for that page which i do not know how to edit:

`faustwp_exclude_from_public_redirect`: Choose the [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) to use when generating redirects when the [enable public route redirects](https://faustjs.org/docs/faustwp/settings#enabling-public-route-redirects) setting is active.

Parameter: $status_code (int): HTTP Status code to use on redirects. Default: `302`

Example Usage
```php
// Make the homepage redirect 301 while all others remain unchanged
add_filter('faustwp_public_redirect_status_code', function ($status_code) {
  if(is_front_page()){
    return 301;
  }
  return $status_code;
}, 10, 1);
```

## Dependant PRs

none

### DIFF
--- a/.changeset/tasty-walls-thank.md
+++ b/.changeset/tasty-walls-thank.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/wordpress-plugin': minor
+---
+
+Added new filter `faustwp_public_redirect_status_code`, allowing WordPress plugins and themes to choose the [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) to use when generating redirects when the [enable public route redirects](https://faustjs.org/docs/faustwp/settings#enabling-public-route-redirects) setting is active.
--- a/plugins/faustwp/includes/deny-public-access/callbacks.php
+++ b/plugins/faustwp/includes/deny-public-access/callbacks.php
@@ -60,7 +60,7 @@ function deny_public_access() {
 	// Get the request uri with query params.
 	$request_uri = home_url( add_query_arg( null, null ) );
 
-	$response_code = 302;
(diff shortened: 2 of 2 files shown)


josephfusco:
Hey @infn8, thank you for taking the time to create this pull request! This looks great to me, and the documentation is clear. I can add it to the main site once a release is cut. 🎉

josephfusco (approved):

infn8:
@josephfusco My pleasure. Glad it works well. I'll keep helping where I can. Thanks for all your work as well.

PR #1995, opened 2024-11-27
[chore]: merge `canary` to `main`

moonmeister:
yup, updating main from canary to remove legacy things


moonmeister:
@josephfusco Why does this have all the issues and commits your PR had? should main be basically up to date?

moonmeister:
@josephfusco I'm wondering if we didn't maintain history and it's causing issues. #1996 seems more like it should be. I'm thinking maybe squash and merge shouldn't be our default? at least not for these big branch changes.

PR #2009, opened 2024-12-12
fix(FeaturedImage): prevent simultaneous width and fill props usage

whoami-pwd:
## Description

This PR fixes the issue where the FeaturedImage component from the example was experiencing conflicts between width and fill properties. The error occurred when both properties were being passed simultaneously.

Changes made:

- Prevented simultaneous usage of width and fill properties
- Added proper property validation
- Updated the layout handling logic to ensure these props are mutually exclusive

## Testing

Local Testing:

`npm run dev`

- Navigate to pages with FeaturedImage components
- Verify images render correctly with 'fill' layout
- Verify images render correctly with explicit width/height

## Screenshots

Screenshot 2024-12-09 at 10 40 46


theodesp:
Awesome!

theodesp (approved):

Issue #1986, opened 2024-11-14
[bug]: Erroneously throwing error


moonmeister:
### Description

We were getting :

`Error: useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider`

This simply wasn't the case. AS you can se from the `_app.js`

```js
import "../../faust.config";
import { WordPressBlocksProvider } from "@faustwp/blocks";
import { FaustProvider } from "@faustwp/core";
import "./global.css";
import Layout from "@/components/layout";
import "@faustwp/core/dist/css/toolbar.css";
import blocks from "@/wp-blocks";

export default function MyApp({ Component, pageProps }) {
	return (
		<FaustProvider pageProps={pageProps}>
			<WordPressBlocksProvider config={{ blocks, theme: null }}>
				<Layout>
					<Component {...pageProps} />
				</Layout>
			</WordPressBlocksProvider>
		</FaustProvider>
	);
}
```

A quick mental check told me I'd recently edited one line in here. Reverting that fixed this issue:

```js
<WordPressBlocksProvider config={{ blocks, theme: undefined }}>
```

This was strange as in JS these should be practically the same.

Some code seleuthing quickly turned up: 

https://github.com/wpengine/faustjs/blob/e17d3188b3aabb1384decc75d4b1b4c217a9b4fb/packages/blocks/src/components/WordPressBlocksProvider.tsx#L75

The use of `===` means we don't type cast `null` to undefined even if it's practically the same. And allowing `null` to work while `undefined` doesn't. 

So, being the docs show a prolific use of `null` to no pass a theme but `undefined` doesn't work, neither does simply not passing the `theme` in the config object. Seems like the fundamental issue here is we're checking for the `WordPressBlocksProvider` based on the value of `theme` and not wether the `WordPressBlocksContext` or `WordPressThemeContext` are actually available

### Steps to reproduce

See above

### Additional context

related: https://github.com/wpengine/faustjs/issues/1844

### @faustwp/core Version

3.1.0

### @faustwp/cli Version

3.1.1

### FaustWP Plugin Version

na

### WordPress Version

na

### Additional environment details

@faustwp/blocks@4.1.0

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

josephfusco:
[...]

PR #2013, opened 2024-12-20
Bug:  Fixing null/undefined theme argument for the WordPressBlocksProvider

colinmurphy:
## Tasks

## Description

Fixed an issue with the theme argument for the WordPressBlockProvider to allow for null or undefined #1986

The condition to check if WordPressThemeContext is not needed as theme is an optional argument as per our docs - https://faustjs.org/reference/wordpressblocksprovider

e.g.

```js
      <WordPressBlocksProvider
        config={{
          blocks,
          theme: null
        }}>
        <Component {...pageProps} key={router.asPath} />
      </WordPressBlocksProvider>
```

So setting the theme argument null, undefined or not at all should not result in an error.

### How to replicate the issue

If you setup an Next.js app with Faust Blocks you should be able to replicate the error.

```bash
cd directory/
npx create-next-app@latest
cd my-app #assuming you set my-app as the project name
npm install @faustwp/blocks
```

**src/pages/_app.js**
```js
import "@/styles/globals.css";
import { WordPressBlocksProvider } from "@faustwp/blocks";
import blocks from "../wp-blocks";


export default function App({ Component, pageProps }) {

  return (
    <WordPressBlocksProvider config={{ blocks, theme: undefined}}>
      <Component {...pageProps} />
    </WordPressBlocksProvider>
  );
}
```

**src/pages/index.js**
```js
import { WordPressBlocksViewer } from "@faustwp/blocks";

export default function Home() {
  return (
    <div>
      <h1>Home</h1>
      <WordPressBlocksViewer blocks={
[
  {
    "name": "core/paragraph",
    "__typename": "CoreParagraph",
    "apiVersion": 3,
    "isDynamic": false,
    "attributes": {
      "cssClassName": null,
      "backgroundColor": null,
      "content": "A JavaScript framework that makes building headless WordPress simple and easy.",
      "style": null,
      "textColor": null,
      "fontSize": null,
      "fontFamily": null,
      "direction": null,
      "dropCap": false,
      "gradient": null,
      "align": null
    }
  },
  {
    "name": "core/paragraph",
    "__typename": "CoreParagraph",
    "apiVersion": 3,
    "isDynamic": false,
    "attributes": {
      "cssClassName": null,
      "backgroundColor": null,
      "content": "A familiar publisher experience, providing a cohesive experience with wp-admin. Easily extensible with built in filters, empowering developers to customize for different use cases.",
      "style": null,
      "textColor": null,
      "fontSize": null,
      "fontFamily": null,
      "direction": null,
      "dropCap": false,
      "gradient": null,
      "align": null
    }
  }
        ]
      } />
    </div>
  );
};

```

**src/wp-blocks/index.js**

```
import { CoreBlocks } from "@faustwp/blocks";

export default {
	...CoreBlocks,
};
```

if you run `npm run dev` you will get the following error:

```
useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider
```

The following PR fixes that issue.

Screenshot 2024-12-24 at 08 57 05

### Further Discussion

After debugging this issue with @moonmeister we also found that the theme argument appears to be redundant.
If you have a look at any of the blocks, it is only passed to the getStyles component which doesn't use the theme argument.

e.g.

https://github.com/wpengine/faustjs/blob/canary/packages/blocks/src/blocks/CoreParagraph.tsx
https://github.com/wpengine/faustjs/blob/canary/packages/blocks/src/utils/get-styles/getStyles.ts
https://github.com/wpengine/faustjs/blob/canary/packages/blocks/src/utils/get-styles/getTextStyles.ts#L10

There is a question on whether this is needed as none of the style components use this argument.

## Related Issue(s):

#1986

## Testing

As described above using a next.js app

## Screenshots

Screenshot 2024-12-20 at 15 17 11

### DIFF
--- a/packages/blocks/src/components/WordPressBlocksProvider.tsx
+++ b/packages/blocks/src/components/WordPressBlocksProvider.tsx
@@ -69,17 +69,19 @@ export function WordPressBlocksProvider(props: {
  * ```
  */
 export function useBlocksTheme(): BlocksTheme {
-
   // If it's an empty object, the provider hasn't been initialized.±
-	if (typeof WordPressBlocksContext === 'undefined' || typeof WordPressBlocksContext === 'undefined') {
-		throw new Error(
+  if (
+    typeof WordPressBlocksContext === 'undefined' ||
+    typeof WordPressBlocksContext === 'undefined'
+  ) {
+    throw new Error(
       'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
     );
-	}
+  }
 
   const themeContext = React.useContext(WordPressThemeContext);
-	if (typeof themeContext === 'undefined') {
-		throw new Error(
+  if (typeof themeContext === 'undefined') {
+    throw new Error(
       'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
     );
   }
--- a/packages/blocks/src/components/WordPressBlocksProvider.tsx
+++ b/packages/blocks/src/components/WordPressBlocksProvider.tsx
@@ -70,6 +70,7 @@ export function WordPressBlocksProvider(props: {
  */
 export function useBlocksTheme(): BlocksTheme {
 
+
   // If it's an empty object, the provider hasn't been initialized.
   if (WordPressBlocksContext == null || WordPressThemeContext == null) {
     throw new Error(
@@ -77,5 +78,12 @@ export function useBlocksTheme(): BlocksTheme {
     );
   }
 
-  return React.useContext(WordPressThemeContext);
+  const themeContext = React.useContext(WordPressThemeContext);
+  if (themeContext == null) {
+    throw new Error(
+      'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
+    );
+  }
+
+  return themeContext;
 }
--- a/packages/blocks/src/components/WordPressBlocksProvider.tsx
+++ b/packages/blocks/src/components/WordPressBlocksProvider.tsx
@@ -68,21 +68,23 @@ export function WordPressBlocksProvider(props: {
  * const theme = useBlocksTheme();
  * ```
  */
-export function useBlocksTheme(): BlocksTheme {
+export function useBlocksTheme(): BlocksTheme | null {
 
-
-  // If it's an empty object, the provider hasn't been initialized.
-  if (WordPressBlocksContext == null || WordPressThemeContext == null) {
-    throw new Error(
+  // If it's an empty object, the provider hasn't been initialized.±
+	if (typeof WordPressBlocksContext === 'undefined') {
+		throw new Error(
       'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
     );
+	}
+
+	// Theme is optional
+	if (typeof WordPressBlocksContext === 'undefined' || WordPressBlocksContext === null) {
+		return null;
   }
 
   const themeContext = React.useContext(WordPressThemeContext);
-  if (themeContext == null) {
-    throw new Error(
-      'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
-    );
+	if (typeof themeContext === 'undefined' || themeContext === null) {
+		return null;
   }
 
   return themeContext;
--- a/packages/blocks/src/components/WordPressBlocksProvider.tsx
+++ b/packages/blocks/src/components/WordPressBlocksProvider.tsx
@@ -69,14 +69,13 @@ export function WordPressBlocksProvider(props: {
  * ```
  */
 export function useBlocksTheme(): BlocksTheme {
-  const themeContext = React.useContext(WordPressThemeContext);
 
   // If it's an empty object, the provider hasn't been initialized.
-  if (themeContext === undefined) {
+  if (WordPressBlocksContext == null || WordPressThemeContext == null) {
     throw new Error(
       'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
     );
   }
 
-  return themeContext;
+  return React.useContext(WordPressThemeContext);;
 }
--- a/packages/blocks/src/components/WordPressBlocksProvider.tsx
+++ b/packages/blocks/src/components/WordPressBlocksProvider.tsx
@@ -77,5 +77,5 @@ export function useBlocksTheme(): BlocksTheme {
     );
   }
 
-  return React.useContext(WordPressThemeContext);;
+  return React.useContext(WordPressThemeContext);
 }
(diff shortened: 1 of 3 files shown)


colinmurphy:
@moonmeister Can you review this please and see if you can still replicate issue #1986  please

colinmurphy:
@moonmeister I need to update this PR as some of the logic isn't correct

colinmurphy:
@moonmeister This is ready for review.

moonmeister:
Can you explain why/how this fixes the issue?

colinmurphy:
@moonmeister  

What PR should solve is the issue around not allowing null for the theme argument for the WordPressBlockProvider.  FWIW I couldn't ever replicate the bug but this is what the PR should do

1. Checks if both WordPressBlocksContext and WordPressThemeContext is defined, if not then it should throw an error (spotted a typo in the code)

2. Then it should allow WordPressThemeContext values to be null. It will only throw an error if undefined

colinmurphy:
@moonmeister 

Thank you for help on debugging this issue. I have updated the PR. I have also updated details on how to replicate the issue and also how it appears theme might be a redundant argument.

packages/blocks/src/components/WordPressBlocksProvider.tsx
```diff
@@ -69,14 +69,5 @@ export function WordPressBlocksProvider(props: {
...
-
-  // If it's an empty object, the provider hasn't been initialized.
-  if (themeContext === undefined) {
-    throw new Error(
-      'useBlocksTheme hook was called outside of context, make sure your app is wrapped with WordPressBlocksProvider',
-    );
-  }
-
-  return themeContext;
+  return React.useContext(WordPressThemeContext) as BlocksTheme;
```
theodesp: I think we should revert useBlocksTheme to its original implementation because it contains a critical check to ensure that an error is triggered when the hook is used outside of the `WordPressBlocksProvider`.

```js
import React from 'react';
import { useBlocksTheme } from '@faustwp/blocks';

function MyComponent() {
  const theme = useBlocksTheme();

  return <div>Current Theme: {JSON.stringify(theme)}</div>;
}

export default function App() {
  return (
    <div>
      {/* Forgot to wrap the app with WordPressBlocksProvider */}
      <MyComponent />
    </div>
  );
}
```

I think the code needs to change is in `WordPressBlocksProvider` so that we conditionally  wrap the content with `WordPressThemeContext` if the theme exists. Something like that:

```js
export function WordPressBlocksProvider(props: {
  children: React.ReactNode;
  config: WordPressBlocksProviderConfig;
}) {
  const { children, config } = props;
  const { blocks, theme } = config;

  const content = (
    <WordPressBlocksContext.Provider value={blocks}>
      {children}
    </WordPressBlocksContext.Provider>
  );

  return theme ? (
    <WordPressThemeContext.Provider value={theme}>
      {content}
    </WordPressThemeContext.Provider>
  ) : (
    content
  );
}
```

moonmeister: This will still throw an error cause we call `useBlocksTheme` in every core block even though, as best I can tell, we don't ever leverage that theme. So while this could be a good fix, it would be incomplete, I believe.

theodesp: We could just remove the `useBlocksTheme` usages inside the blocks. Or maybe even better, always assign a default value for theme as `{}` so we won't have to change anything else.

```
export function WordPressBlocksProvider(props: {
  children: React.ReactNode;
  config: WordPressBlocksProviderConfig;
}) {
  const { children, config } = props;
  const { blocks, theme = {} } = config;

  return (
    <WordPressThemeContext.Provider value={theme}>
      <WordPressBlocksContext.Provider value={blocks}>
        {children}
      </WordPressBlocksContext.Provider>
    </WordPressThemeContext.Provider>
  );
}
```

colinmurphy: @theodesp I might be wrong here but would we also need to then update the [getStyles](packages/blocks/src/utils/get-styles/getStyles.ts) function and style blocks (e.g. [getBorderStyles](https://github.com/wpengine/faustjs/blob/canary/packages/blocks/src/utils/get-styles/getBorderStyles.ts#L10) as `useBlocks` is used as the theme argument in all core blocks for this function.

colinmurphy: > We could just remove the `useBlocksTheme` usages inside the blocks. Or maybe even better, always assign a default value for theme as `{}` so we won't have to change anything else.
> 
> ```
> export function WordPressBlocksProvider(props: {
>   children: React.ReactNode;
>   config: WordPressBlocksProviderConfig;
> }) {
>   const { children, config } = props;
>   const { blocks, theme = {} } = config;
> 
>   return (
>     <WordPressThemeContext.Provider value={theme}>
>       <WordPressBlocksContext.Provider value={blocks}>
>         {children}
>       </WordPressBlocksContext.Provider>
>     </WordPressThemeContext.Provider>
>   );
> }
> ```

Thanks @theodesp this does work setting a default for the theme so I am just going to update the PR shortly.

packages/blocks/tests/components/WordPressBlocksProvider.test.tsx
moonmeister, line 1: @colinmurphy I'm guessing we need to leave these tests in now that we've changed the fix, yes?

colinmurphy: @moonmeister We probably don't. I was leaving this for reference if someone wanted to use the theme argument but I removed the failing test for undefined.

colinmurphy:
Just confirming I did some further testing and the following 3 scenarios worked. I just wanted to test again as we will be updating the docs for Faust to remove the theme argument.

```js
// src/pages/_app.js
import "@/styles/globals.css";
import { WordPressBlocksProvider } from "@faustwp/blocks";
import blocks from "../wp-blocks";
export default function App({ Component, pageProps }) {

  return (
    <WordPressBlocksProvider config={{ blocks}}>
      <Component {...pageProps} />
    </WordPressBlocksProvider>
  );
}
```

```js
// src/pages/_app.js
import "@/styles/globals.css";
import { WordPressBlocksProvider } from "@faustwp/blocks";
import blocks from "../wp-blocks";
export default function App({ Component, pageProps }) {

  return (
    <WordPressBlocksProvider config={{ blocks, theme: null}}>
      <Component {...pageProps} />
    </WordPressBlocksProvider>
  );
}
```


```js
// src/pages/_app.js
import "@/styles/globals.css";
import { WordPressBlocksProvider } from "@faustwp/blocks";
import blocks from "../wp-blocks";
export default function App({ Component, pageProps }) {

  return (
    <WordPressBlocksProvider config={{ blocks, theme: undefined}}>
      <Component {...pageProps} />
    </WordPressBlocksProvider>
  );
}
```

theodesp (approved):

packages/blocks/tests/components/WordPressBlocksProvider.test.tsx
colinmurphy: @moonmeister Tests reverted back. Thanks for pointing this out.

josephfusco (approved):
LGTM!

PR #2046, opened 2025-02-18
docs: transfer new docs

moonmeister:
I'm copying the documents here. The lift to move the full site is going to prove too much work at the moment. These will be pulled into the site from here.

Leaving it as a draft till we get the last couple merged.


moonmeister:
This is ready for review. I don't think there's much to review, as most of it is copied and pasted from the other repo. I changed file extensions from `mdx` => `md` and changed from a JS `export const metadata` to Frontmatter. That should be all done. 

these docs live on [PR Preview](https://hicuy553ynd6o7orxs95na6l4.js.wpenginepowered.com/)

colinmurphy (approved):
@moonmeister LGTM ❤️ 🚀 

I manually checked random files against the .org repo and checked it locally and the files matched. I did notice these are md rather than mdx but more noting it than anything.

We should be fully able to test once merged in the .org PR

PR #2180, opened 2025-09-23
[next]: add existing Faust utilities as a separate package 

### LINKED ISSUE #2152: [experimental]: Add existing Faust utils to the new Faust
Copy existing Faust utils into the new Faust and implement to existing packages.

ahuseyn:
## Description
This pull request introduces a new `@faustjs/utilities` package, providing the existing utility functions of Faust.js. The changes include several utility modules, configuration for building both ESM and CommonJS outputs, and documentation for usage and contribution.

**Package Setup and Documentation:**

* Added `package.json` for the new `@faustjs/utilities` package, specifying dependencies (`chalk`, `lodash`), build scripts, and engine requirements.
* Added a comprehensive `README.md` describing installation, available utilities, usage examples, and contribution guidelines.

## Related Issue(s):

- #2152

### DIFF
--- a/packages/utilities/package.json
+++ b/packages/utilities/package.json
@@ -1,0 +1,25 @@
+{
+	"name": "@faustjs/utilities",
+	"version": "0.0.0",
+	"description": "A collection of utility functions for Faust.js",
+	"main": "dist/cjs/index.js",
+	"module": "dist/mjs/index.js",
+	"types": "dist/mjs/index.d.ts",
+	"devDependencies": {
+		"@types/node": "^18.0.6"
+	},
+	"dependencies": {
+		"chalk": "^4.1.2",
+		"lodash": "^4.17.21"
+	},
+	"scripts": {
+		"build": "npm run build-esm && npm run build-cjs",
+		"build-esm": "tsc -p .",
+		"build-cjs": "tsc -p tsconfig.cjs.json"
+	},
+	"author": "Faust.js Team",
+	"engines": {
+		"node": ">=18",
+		"npm": ">=8"
+	}
+}
--- a/packages/utilities/package.json
+++ b/packages/utilities/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@faustjs/utilities",
-	"version": "0.0.0",
+	"version": "0.0.1",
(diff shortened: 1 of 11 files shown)


packages/utilities/package.json
```diff
@@ -0,0 +1,25 @@
+{
+	"name": "@faustjs/utilities",
+	"version": "0.0.0",
```
josephfusco: We _might_ need an initial version here other than 0.0.0

packages/utilities/src/assert.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export function isValidUrl(url: string): boolean {
+	try {
+		// eslint-disable-next-line no-new
+		new URL(url);
+		return true;
+	} catch (e) {
+		return false;
+	}
+}
```
josephfusco: I think we can simplify this by reusing [isURL](https://github.com/WordPress/gutenberg/tree/trunk/packages/url/src/is-url.ts) from @wordpress/url instead of maintaining a duplicate helper.

Using the shared utility keeps our implementation consistent with Gutenberg and removes some maintenance overhead.  What do you think about switching to that here?

packages/utilities/src/assert.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Returns whether or not a string is a valid email address
+ *
+ * @export
+ * @returns
+ */
+export function isValidEmail(email: string): boolean {
+	return emailRegex.test(String(email).toLowerCase());
+}
```
josephfusco: We could replace this helper with [isEmail](https://github.com/WordPress/gutenberg/blob/trunk/packages/url/src/is-email.ts) from @wordpress/url instead of maintaining our own regex.  This keeps us consistent with Gutenberg and reduces maintenance overhead.

josephfusco (changes requested):

packages/utilities/src/assert.ts
ahuseyn: Thanks @josephfusco, fixed this and the other issues above.

josephfusco (approved):
This looks good to me! Nice work @ahuseyn

PR #2188, opened 2025-10-02
feat(faustwp-core): add typed GraphQL client (fetchGraphQL<T>) and FaustClient

ArjunSaluja:
Adds a typed GraphQL client and helper to `packages/faustwp-core`.
.fetchGraphQL<T>` — a generic typed fetch wrapper for GraphQL.
.`FaustClient` — lightweight client for queries and mutations.
. Shared GraphQL response types (`types.ts`).
. Includes a small Jest test for typed usage.

.Improves developer experience and ensures GraphQL responses are strongly typed.

Example
```ts
type Result = { post: { id: string; title: string } };
const client = new FaustClient(process.env.FAUST_GRAPHQL_URL!);
const { post } = await client.query<Result>(`query { post(id: "1") { id title } }`);


theodesp:
Hi @ArjunSaluja 
Thank you for your effort.
This appears to be a breaking change, as it removes the Apollo Client implementation in favor of a custom FaustClient and direct fetch-based GraphQL requests. 

We do not aim to remove Apollo Client support from Faust at this time.

Can you share more about your use case? Is there something you’re trying to achieve that the current Apollo Client integration does not support? 
We’d love to understand your requirements and see if there’s a way to address your needs without removing existing features.

Thanks!

ArjunSaluja:
Hey, thanks for the feedback!

My intention wasn’t to remove Apollo Client or introduce any breaking
change. I was mainly trying to add a lightweight, type-safe helper (
fetchGraphQL<T> and FaustClient) as an *optional utility*, not as a
replacement.

The idea was to make it easier to:

   -

   Run quick GraphQL queries in scripts or tests without setting up Apollo
   -

   Provide a simpler example for people exploring Faust.js
   -

   Have a smaller, generic option for cases like CLI tools or build-time
   fetching

Totally understand that Apollo is core to Faust — I can definitely adjust
this PR so it keeps Apollo intact. Maybe this could live under something
like utils/fetchGraphQL.ts or be marked experimental if that’s better.

Let me know what you think — happy to update it to fit within the existing
setup.

theodesp:
@ArjunSaluja 

Thank you for sharing your perspective and proposal!

We appreciate your effort to add a lightweight, type-safe GraphQL helper as an optional utility.

However, Faust.js is tightly integrated with Apollo Client, and most users rely on its features and patterns. Typing Apollo queries is straightforward, as developers have direct access to the Apollo Client itself; Faust does not interfere or abstract away Apollo’s API. 

If someone prefers a custom abstraction like FaustClient, they can easily implement that in their own project as needed.

Given this, we’re unlikely to add this feature to the Faust core package at this time. We want to avoid expanding the API surface with utilities that may duplicate what’s already available or introduce confusion around recommended usage patterns.

Additionally, we’ve just published our [Q4 roadmap](https://github.com/wpengine/faustjs/discussions/2195),  which includes fundamental changes and improvements to Faust in the next major version. We’re focusing on core architecture and features, so we’re cautious about introducing new utilities outside that scope for now.

Thank you again for the suggestion and for your engagement! 
If you have other ideas, or want to discuss how best to achieve type-safe GraphQL queries with Faust/Apollo, we’re happy to help.

ArjunSaluja:
Thanks for your explanation.

Issue #2275, opened 2026-02-23
SPIKE: Investigate Dynamic Imports Implementation


ahuseyn:
### Description

The Dynamic Imports feature implemented recently shown some issues with the Getting Started Example using Nextjs v16. These issues include:
1. Components rendering twice if no "loading" option provided or client-only loading provided
2. Hydration error (client/server mismatch) if "loading" option provided
Using "ssr:false" fixes the issue, but it skips server rendering.
The root cause of the issue should be investigated.

### Steps to reproduce

1. Clone the repo
2. Install dependencies
3. Run `npm run dev`
4. Update wp-templates/index.js file with the provided combinations above to test different scenarios

### @faustwp/core Version

3.3.6

### @faustwp/cli Version

3.3.6

### FaustWP Plugin Version

https://faustexample.wpengine.com

### WordPress Version

https://faustexample.wpengine.com

### Additional environment details

.env

```
# Your WordPress site URL
NEXT_PUBLIC_WORDPRESS_URL=https://faustexample.wpengine.com
```

### Please confirm that you have searched existing issues in the repo.

- [x] Yes

ahuseyn:
## Problem

`WordPressTemplateInternal` passed `unknownTemplate` directly to `React.createElement` for dynamic templates. Since `unknownTemplate` is a `next/dynamic()` wrapper, React was rendering the wrapper itself rather than the actual component function:

```tsx
// rendering the next/dynamic() wrapper directly
const Component = unknownTemplate as React.FC<{ [key: string]: any }>;
return React.createElement(Component, newProps, null);
```

The `next/dynamic()` wrapper has its own internal loading state and lifecycle. This caused two issues:

- **Hydration mismatch** — the wrapper renders a `loading` fallback on the client before the JS chunk loads, which differs from what the server rendered.
- **Double render** — the wrapper's internal resolution competed with Faust's own data-fetching `useEffect`, causing the template to render twice.

## Fix

Resolve the dynamic wrapper to its inner component via `loadDynamicComponent` and store it in a dedicated `resolvedTemplate` state. Render `null` until resolved, then render the actual inner component directly, bypassing the `next/dynamic` wrapper entirely.

```tsx
const isDynamic = isDynamicComponent(unknownTemplate);
const [resolvedTemplate, setResolvedTemplate] =
    useState<WordPressTemplateType | null>(null);

useEffect(() => {
    if (!unknownTemplate || !isDynamic) {
        return;
    }
[...]

PR #2296, opened 2026-03-23
fix: Resolve dynamic templates to inner component before rendering to…

ahuseyn:
This PR fixes an issue where dynamic template components were rendered via the next/dynamic wrapper directly, causing hydration mismatches and double renders, by resolving the dynamic component to its inner function and storing it in state before rendering.

## Description

Previously, `WordPressTemplateInternal` passed `unknownTemplate` directly to `React.createElement` when rendering dynamic templates. Since `unknownTemplate` in this case is a `next/dynamic()` wrapper (not the actual component function) React was rendering the wrapper itself, which has its own internal loading state and lifecycle. This caused the component to flash its `loading` fallback on the client during hydration, and in some cases trigger double renders as the wrapper's internal resolution competed with Faust's own data-fetching effects.

The fix resolves the dynamic wrapper to its actual inner component via `loadDynamicComponent` and stores it in a separate `resolvedTemplate` state using a function updater (`() => template`) to prevent React from mistakenly invoking the component as a state initializer. 

## Related Issues

https://github.com/wpengine/faustjs/issues/2275

### DIFF
--- a/.changeset/shaky-toes-open.md
+++ b/.changeset/shaky-toes-open.md
@@ -1,0 +1,5 @@
+---
+'@faustwp/core': patch
+---
+
+Fixed an issue where dynamic template components were rendered via the next/dynamic wrapper directly, causing hydration mismatches and double renders, by resolving the dynamic component to its inner function and storing it in state before rendering.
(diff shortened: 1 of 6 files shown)


josephfusco (approved):
I'm not getting any hydration errors in the console, errors/warnings are pre-existing to this PR.  Nice work!


Image

PR #2312, opened 2026-04-03
fix[faustwp]: (#2310) use hash_equals for secret key comparison

### LINKED ISSUE #2310: fix[faustwp]: secret key comparisons use timing-vulnerable === instead of hash_equals()
### Description

The FaustWP plugin authenticates REST API and GraphQL requests by comparing the shared secret key from the `x-faustwp-secret` (or `x-wpe-headless-secret`) HTTP header against the stored value. Three locations use PHP's `===` operator for this comparison:

- `rest/callbacks.php:422` — `rest_authorize_permission_callback()` (guards all `faustwp/v1/*` REST endpoints)
- `rest/callbacks.php:447` — `wpac_authorize_permission_callback()` (guards legacy `wpac/v1/authorize` endpoint)
- `graphql/callbacks.php:70` — `filter_introspection()` (controls GraphQL introspection access)

PHP's `===` compares byte-by-byte and short-circuits on the first mismatch. `hash_equals()` performs constant-time comparison regardless of where the mismatch occurs.

The codebase already uses `hash_equals()` for HMAC validation in `auth/functions.php:200`, so these three spots are an inconsistency.

### Practical Impact

Remote timing attacks over network are difficult due to jitter, but `hash_equals()` is a PHP security best practice (OWASP recommendation) and eliminates the vector entirely.

### CWE Classification

CWE-208: Observable Timing Discrepancy

### Environment

- FaustWP plugin v1.8.6
- `canary` branch

### LINKED ISSUE #2311: fix[faustwp]: uploaded blockset file not cleaned up when extraction fails
### Description

In `plugins/faustwp/includes/blocks/functions.php`, the `process_and_replace_blocks()` function moves an uploaded file to the target directory, then attempts to unzip it. If `unzip_uploaded_file()` fails, the function returns the `WP_Error` immediately without deleting the moved file:

```php
$move_result = move_uploaded_file( $wp_filesystem, $file, $target_file );
if ( is_wp_error( $move_result ) ) {
    return $move_result;
}

$unzip_result = unzip_uploaded_file( $target_file, $dirs['target'] );
if ( is_wp_error( $unzip_result ) ) {
    return $unzip_result;  // $target_file left on disk
}
```

The orphaned file remains at a predictable path under `wp-content/uploads/faustwp/blocks/`. While the endpoint requires secret key authentication, leftover files accumulate on repeated failed uploads and are not cleaned up.

### Expected Behavior

The uploaded file should be deleted from the target directory before returning the error, so failed uploads don't leave artifacts on disk.

### CWE Classification

CWE-459: Incomplete Cleanup

### Environment

- FaustWP plugin v1.8.6
- `canary` branch

latenighthackathon:
## Summary

The shared secret key (`x-faustwp-secret` header, plus the deprecated `x-wpe-headless-secret` and the GraphQL `HTTP_X_FAUST_SECRET` server header) was compared using PHP's `===` / `!==` operators in three permission callbacks. Those operators short-circuit on the first byte mismatch, making them theoretically vulnerable to timing side-channel attacks (CWE-208). The codebase already uses `hash_equals()` for HMAC validation in `auth/functions.php:200` — these three call sites were inconsistent.

This PR swaps all three to `hash_equals()` for constant-time comparison, and (for the GraphQL site) adds the WordPress-coding-standards-recommended `sanitize_text_field( wp_unslash( ... ) )` on the raw `$_SERVER` read before passing to `hash_equals()`.

## Changes

- `plugins/faustwp/includes/rest/callbacks.php` — `rest_authorize_permission_callback()` and `wpac_authorize_permission_callback()`: `=== $header_key` → `hash_equals( $secret_key, $header_key )`.
- `plugins/faustwp/includes/graphql/callbacks.php` — `filter_introspection()`: `!== $_SERVER['HTTP_X_FAUST_SECRET']` → `! hash_equals( $secret_key, $faust_secret )` after `sanitize_text_field( wp_unslash( ... ) )`.

## Related Issue

Closes #2310.

(The blockset cleanup that originally rode along here was split out per @josephfusco's review feedback and moved to #2338.)

## Confirm the issue (before the fix)

```bash
git checkout canary
grep -n '=== $header_key' plugins/faustwp/includes/rest/callbacks.php
# Expected: matches at lines 422 and 447

grep -n "!== \$_SERVER\['HTTP_X_FAUST_SECRET'\]" plugins/faustwp/includes/graphql/callbacks.php
# Expected: match at line 70
```

## Confirm the fix

```bash
grep -n 'hash_equals' plugins/faustwp/includes/rest/callbacks.php plugins/faustwp/includes/graphql/callbacks.php
# Expected: rest/callbacks.php two matches; graphql/callbacks.php one match.

grep -B 1 'hash_equals' plugins/faustwp/includes/graphql/callbacks.php
# Expected: sanitize_text_field( wp_unslash( ... ) ) on the preceding line.
```

## Run the test suite

> NOTE: there is no `package.json` in `plugins/faustwp/`. The plugin's scripts are Composer scripts, not npm scripts. The commands below mirror what CI runs (`.github/workflows/unit-test-plugin.yml`).

```bash
cd plugins/faustwp
composer install
composer phpcs

# Bring up the test stack
WP_VERSION=6.8 composer run docker:start
docker compose exec wordpress init-testing-environment.sh
docker compose exec -w /var/www/html/wp-content/plugins/faustwp wordpress \
  wp plugin install wp-graphql --activate --allow-root

# The four pre-existing filter_introspection tests in GraphQLCallbacksTests
# already cover the GraphQL change. This PR adds RestCallbacksTests.php
# (previously zero coverage for the REST permission callbacks).
docker compose exec -w /var/www/html/wp-content/plugins/faustwp wordpress \
  ./vendor/bin/phpunit --filter='RestCallbacks|filter_introspection'
# Expected: OK (12 tests, 14+ assertions)

# Full suite (single-site + multisite), what CI runs
docker compose exec -w /var/www/html/wp-content/plugins/faustwp wordpress \
  composer test
# Expected: OK on both runs
```

The `test_secret_comparisons_use_constant_time_hash_equals` test is the regression guard against a revert to `===` / `!==`. Behavioural tests alone can't distinguish `hash_equals` from `===` for valid/invalid inputs (both return the same boolean; the difference is timing), so the guard narrowly checks the source for the known-bad shapes this PR removes.

## Verified red-on-revert

Manually reverted `hash_equals` → `===` / `!==` in both files locally; the `test_secret_comparisons_use_constant_time_hash_equals` test fails cleanly with PHPUnit's `Failed asserting that ... does not contain "=== $header_key"`. Restoring `hash_equals` greens the suite again.

## Checklist

- [x] Targets `canary`
- [x] Changeset added (`@faustwp/wordpress-plugin` patch)
- [x] Regression tests added (`tests/integration/RestCallbacksTests.php`) — covers both `rest_authorize_permission_callback` and the deprecated `wpac_authorize_permission_callback`, plus the source-level guard
- [x] PHPCS clean on touched files
- [x] PHPUnit green on single-site and multisite (129 tests, 288 assertions on this branch)

### DIFF
--- a/.changeset/hash-equals-secret-comparison.md
+++ b/.changeset/hash-equals-secret-comparison.md
@@ -1,0 +1,5 @@
+---
+"@faustwp/wordpress-plugin": patch
+---
+
+fix[faustwp]: use hash_equals() for constant-time secret key comparison in REST and GraphQL permission callbacks
--- a/plugins/faustwp/includes/graphql/callbacks.php
+++ b/plugins/faustwp/includes/graphql/callbacks.php
@@ -66,8 +66,9 @@ function filter_introspection( $value, $default_value, $option_name, $section_fi
 		return $value;
 	}
 
-	$secret_key = get_secret_key();
-	if ( $secret_key !== $_SERVER['HTTP_X_FAUST_SECRET'] ) {
+	$secret_key   = get_secret_key();
+	$faust_secret = sanitize_text_field( wp_unslash( $_SERVER['HTTP_X_FAUST_SECRET'] ) );
+	if ( ! hash_equals( $secret_key, $faust_secret ) ) {
 		return $value;
 	}
 
--- a/plugins/faustwp/includes/rest/callbacks.php
+++ b/plugins/faustwp/includes/rest/callbacks.php
@@ -419,7 +419,7 @@ function rest_authorize_permission_callback( \WP_REST_Request $request ) {
 	$header_key = $request->get_header( 'x-faustwp-secret' );
 
 	if ( $secret_key && $header_key ) {
-		return $secret_key === $header_key;
+		return hash_equals( $secret_key, $header_key );
 	}
 
 	return false;
@@ -444,7 +444,7 @@ function wpac_authorize_permission_callback( \WP_REST_Request $request ) {
 	$header_key = $request->get_header( 'x-wpe-headless-secret' );
 
 	if ( $secret_key && $header_key ) {
-		return $secret_key === $header_key;
+		return hash_equals( $secret_key, $header_key );
 	}
 
 	return false;


Fran-A-Dev (commented):
The REST calls look ok. The GraphQL path might need one other guard.  `$_SERVER['HTTP_X_FAUST_SECRET']` is sanitized before comparison, but it is still accessed directly without an `isset()` check. If that header is missing, this could still trigger an undefined index notice. @moonmeister @latenighthackathon do we default to an empty string  before calling `hash_equals()` so the comparison fails cleanly when the header is absent?

latenighthackathon:
Good catch to flag this - the existing `isset()` guard on line 65 already handles this case:

```php
if ( ! isset( $_SERVER['HTTP_X_FAUST_SECRET'] ) ) {
    return $value;
}
```

This should return early before we ever reach the `sanitize_text_field( wp_unslash( $_SERVER['HTTP_X_FAUST_SECRET'] ) )` call on line 70, so there's no risk of an undefined index notice when the header is absent. That guard was already in the upstream code - our diff just changed the comparison below it from `!==` to `hash_equals()` and added the sanitization.

josephfusco (changes requested):
@latenighthackathon Thanks for the fixes. Both changes look correct.

Can we split this into two separate PRs? The `hash_equals()` hardening and the blockset cleanup are unrelated, so separate PRs with separate changesets would make them easier to review and revert independently.

latenighthackathon:
@josephfusco — split done. This PR is now scoped to just the `hash_equals()` hardening (#2310); the blockset cleanup (#2311) moved to #2338. Each has its own changeset. Cheers!

josephfusco:
Closing since this has been split.  Thanks @latenighthackathon!

josephfusco (approved):

josephfusco:
Superseded by #2362 — same security fix (cherry-picked from this PR's `7660c3a4`) plus a regression test for the REST permission callbacks that previously had no coverage. Pushed to a maintainer-controlled branch on wpengine/faustjs because `maintainerCanModify` is false on this PR, which prevented me from pushing the test commit here directly.

Credit for the security work is preserved via Co-authored-by on the new PR. Once #2362 lands, this PR can be closed.

PR #2314, opened 2026-04-03
fix[faustwp-core]: (#2313) add security flags to removeCookie()

### LINKED ISSUE #2313: fix[faustwp-core]: removeCookie() missing path, sameSite, secure, and httpOnly flags
### Description

In `packages/faustwp-core/src/server/auth/cookie.ts`, the `removeCookie()` method expires a cookie with only the `expires` attribute set:

```typescript
cookie.serialize(key, '', {
    expires: new Date(0),
})
```

However, `setCookie()` (called via `setRefreshToken()` in `token.ts`) sets the cookie with full security attributes:

```typescript
{
    path: '/',
    sameSite: 'strict',
    secure: true,
    httpOnly: true,
}
```

Per RFC 6265, a `Set-Cookie` header without a matching `path` defaults to the current request path. If the removal request comes from a path other than `/`, the browser won't delete the original cookie set with `path: '/'`. The cookie persists after what the application considers a successful logout.

Additionally, without `secure` and `httpOnly` on the removal header, there is a brief window where a non-secure cookie with the same name could be injected over HTTP before the expiry takes effect.

### Expected Behavior

`removeCookie()` should include the same `path`, `sameSite`, `secure`, and `httpOnly` flags used when the cookie was set, ensuring the browser correctly identifies and expires the target cookie.

### Environment

- @faustwp/core v3.4.0
- `canary` branch

latenighthackathon:
## Description

`removeCookie()` in `packages/faustwp-core/src/server/auth/cookie.ts` expires the refresh token cookie with only `expires: new Date(0)`, missing the `path`, `sameSite`, `secure`, and `httpOnly` flags that `setCookie()` uses when setting it.

Per RFC 6265, a `Set-Cookie` header without a `path` attribute defaults to the current request path. If the logout request comes from a different path than `/`, the browser won't match and delete the original cookie — it persists after what the application considers a successful logout.

### Before

```typescript
cookie.serialize(key, '', {
    expires: new Date(0),
})
```

### After

```typescript
cookie.serialize(key, '', {
    path: '/',
    sameSite: 'strict',
    secure: true,
    httpOnly: true,
    expires: new Date(0),
})
```

These flags match the attributes used in `setRefreshToken()` (`token.ts:51-58`).

## Related Issues

Closes #2313

**1. Verify `removeCookie()` is missing security flags:**
```bash
sed -n '67,76p' packages/faustwp-core/src/server/auth/cookie.ts
# Expected on canary: only "expires: new Date(0)" — no path, sameSite, secure, or httpOnly
```

**2. Compare with `setCookie()` which has the correct flags:**
```bash
grep -A 7 'this.cookies.setCookie' packages/faustwp-core/src/server/auth/token.ts
# Expected: path: '/', sameSite: 'strict', secure: true, httpOnly: true
# These are the flags that removeCookie() should also have
```

### Confirm the fix

**3. Verify `removeCookie()` now includes all security flags:**
```bash
sed -n '67,80p' packages/faustwp-core/src/server/auth/cookie.ts
# Expected: path, sameSite, secure, httpOnly all present before expires
```

**4. Verify flags match the setCookie() call in token.ts:**
```bash
# Both should use: path: '/', sameSite: 'strict', secure: true, httpOnly: true
diff <(grep -A 4 "path:" packages/faustwp-core/src/server/auth/cookie.ts) \
     <(grep -A 4 "path:" packages/faustwp-core/src/server/auth/token.ts)
# Expected: identical flag values (only expires/maxAge will differ)
```

### Run the test suite

**JS tests** — confirms no regressions:
```bash
npm install && npm run build && npm run test
```

### DIFF
--- a/.changeset/cookie-removal-flags.md
+++ b/.changeset/cookie-removal-flags.md
@@ -1,0 +1,5 @@
+---
+"@faustwp/core": patch
+---
+
+fix[faustwp-core]: add path, sameSite, secure, and httpOnly flags to removeCookie() to match setCookie() attributes
--- a/packages/faustwp-core/src/server/auth/cookie.ts
+++ b/packages/faustwp-core/src/server/auth/cookie.ts
@@ -70,6 +70,10 @@ export class Cookies {
 		this.response?.setHeader(
 			'Set-Cookie',
 			cookie.serialize(key, '', {
+				path: '/',
+				sameSite: 'strict',
+				secure: true,
+				httpOnly: true,
 				expires: new Date(0),
 			}),
 		);


Fran-A-Dev (approved):
LGTM, I don't think the original cookie is ever set with domain attribute anywhere else. This looks good.

PR #1897, opened 2024-05-30
Adds wpAdminUrl hook

theodesp:
## Tasks

- [x] I have signed a [Contributor License Agreement (CLA)](https://github.com/wpengine/faustjs#contributor-license-agreement) with WP Engine.
- [x] If a code change, I have written testing instructions that the whole team & outside contributors can understand.
- [x] I have written and included a comprehensive [changeset](https://github.com/wpengine/faustjs/blob/canary/DEVELOPMENT.md#deployment) to properly document the changes I've made.

## Description

This PR adds a new hook called `wpAdminUrl` that lets users alter the wp-admin url.

## Related Issue(s):

https://github.com/wpengine/faustjs/issues/1882

## Testing

1. Add the following plugin hook into the example project:

```faust.config.js
import { setConfig } from '@faustwp/core';
import templates from './wp-templates';
import possibleTypes from './possibleTypes.json';

import { FaustHooks, FaustPlugin } from '@faustwp/core';

export class MyPlugin {
  apply(hooks) {
    const { addAction, addFilter } = hooks;

    addFilter(
      'wpAdminUrl',
      'my-namespace',
      (wpAdminUrl, context) => {
        console.debug('wpAdminUrl', wpAdminUrl);
        return wpAdminUrl.replace('/wp-admin', '/wp/wp-admin')
      },
    );
  }
}
/**
 * @type {import('@faustwp/core').FaustConfig}
 **/
export default setConfig({
  templates,
  plugins: [new MyPlugin()],
  experimentalToolbar: true,
  possibleTypes,
});

```

2. Open the example project site and review the toolbar url. It should show the updated path:

```
/wp/wp-admin/admin.php?page=graphiql-ide
```

## Screenshots
Screenshot 2024-05-30 at 15 29 24

## Documentation Changes

Needs to add entry in https://faustjs.org/reference/faust-plugin-system-filters

PR #2315, opened 2026-04-03
fix[faustwp]: (#1872) add filter for search_pattern in handle_generate_endpoint

### LINKED ISSUE #1872: Add filter for search_pattern in handle_generate_endpoint
### Context:
I'm testing Faust with WordPress Bedrock, and due to the different structure, when previewing content, it results in a 404 error.

This issue is caused by the `$search_pattern` in `handle_generate_endpoint()`.
https://github.com/wpengine/faustjs/blob/002687fafd0589d718c3d4fa3403b9ff2a3bb88c/plugins/faustwp/includes/auth/callbacks.php#L24-L29

Because Bedrock returns the prefix `/wp/` in `site_url( '/generate', 'relative' )` (`/wp/generate`) while `$_SERVER['REQUEST_URI']` contains only `/generate[...]`, it always fails at the statement below:

```php
if ( ! preg_match( $search_pattern, $_SERVER['REQUEST_URI'] ) ) { // phpcs:ignore WordPress.Security
	return;
}
``` 

### Suggestion
Since `site_url()` returns the URL for the current site where WordPress applications are accessible, the best way to move forward is to use a callback filter (`apply_filters`) so anyone can customize it with a callback filter.

```php
/**
* Filter 'faustwp_generate_endpoint_search_pattern'.
*
* @param string  $value   The search pattern.
*/
$search_pattern = apply_filters( 'faustwp_generate_endpoint_search_pattern', $search_pattern );
```

### Example of how to use the filter
```php
add_filter('faustwp_generate_endpoint_search_pattern', function ($value) {
	return str_replace('/wp/', '/', $value);
});
```

latenighthackathon:
## Description

Non-standard WordPress installations like [Bedrock](https://roots.io/bedrock/) use a different directory structure where `site_url()` returns a path with a `/wp/` prefix (e.g. `/wp/generate`) but `$_SERVER['REQUEST_URI']` contains only `/generate`. This causes the `preg_match` in `handle_generate_endpoint()` to fail, resulting in 404 errors when previewing content.

This PR adds an `apply_filters()` call on `$search_pattern` so sites with non-standard directory structures can adjust the matching pattern without modifying the plugin source.

### Usage example for Bedrock:

```php
add_filter( 'faustwp_generate_endpoint_search_pattern', function ( $pattern ) {
    return str_replace( '/wp/', '/', $pattern );
} );
```

## Related Issues

Closes #1872

**1. Verify no filter exists on `$search_pattern`:**
```bash
grep -n 'apply_filters.*search_pattern' plugins/faustwp/includes/auth/callbacks.php
# Expected on canary: no output
```

**2. Verify `$search_pattern` uses `site_url()` which includes subdirectories on Bedrock:**
```bash
sed -n '25,27p' plugins/faustwp/includes/auth/callbacks.php
# Expected: $search_pattern = ':^' . site_url( '/generate', 'relative' ) . ':';
# On Bedrock, site_url() returns /wp/generate, but REQUEST_URI has /generate
```

### Confirm the fix

**3. Verify the filter is applied after `$search_pattern` is constructed:**
```bash
grep -A 12 "search_pattern = ':^'" plugins/faustwp/includes/auth/callbacks.php | head -15
# Expected: apply_filters( 'faustwp_generate_endpoint_search_pattern', $search_pattern )
# between the assignment and the preg_match
```

**4. Verify the docblock follows WordPress standards (@since, @param):**
```bash
grep -B 10 'apply_filters' plugins/faustwp/includes/auth/callbacks.php | grep '@'
# Expected: @since 1.8.7 and @param string $search_pattern
```

### Run the test suites

**phpcs** — confirms the filter follows WordPress coding standards:
```bash
cd plugins/faustwp && composer install && composer phpcs
```

**JS tests** — confirms no regressions:
```bash
npm install && npm run build && npm run test
```

**WordPress PHPUnit** — confirms plugin functionality:
```bash
cd plugins/faustwp
composer run docker:start
docker-compose exec -e COVERAGE=1 wordpress init-testing-environment.sh
docker exec --workdir=/var/www/html/wp-content/plugins/faustwp $(docker-compose ps -q wordpress) wp plugin install wp-graphql --activate --allow-root
docker exec --workdir=/var/www/html/wp-content/plugins/faustwp $(docker-compose ps -q wordpress) composer test
composer run docker:stop
```

### DIFF
--- a/.changeset/filter-generate-endpoint.md
+++ b/.changeset/filter-generate-endpoint.md
@@ -1,0 +1,5 @@
+---
+"@faustwp/wordpress-plugin": patch
+---
+
+fix[faustwp]: add `faustwp_generate_endpoint_search_pattern` filter to handle_generate_endpoint() for non-standard WordPress installations
--- a/plugins/faustwp/includes/auth/callbacks.php
+++ b/plugins/faustwp/includes/auth/callbacks.php
@@ -23,6 +23,17 @@ add_action( 'parse_request', __NAMESPACE__ . '\\handle_generate_endpoint' );
  */
 function handle_generate_endpoint() {
 	$search_pattern = ':^' . site_url( '/generate', 'relative' ) . ':';
+	/**
+	 * Filter the search pattern used to match the generate endpoint.
+	 *
+	 * Useful for non-standard WordPress installations (e.g. Bedrock) where
+	 * site_url() includes a subdirectory that does not appear in REQUEST_URI.
+	 *
+	 * @since 1.8.7
+	 *
+	 * @param string $search_pattern The regex pattern used to match the generate endpoint.
+	 */
+	$search_pattern = apply_filters( 'faustwp_generate_endpoint_search_pattern', $search_pattern );
 
 	if ( ! preg_match( $search_pattern, $_SERVER['REQUEST_URI'] ) ) { // phpcs:ignore WordPress.Security
 		return;


josephfusco:
Thanks for the contribution!

I don't think we need a new filter here. The fix is to use `home_url()` instead of `site_url()`:

```php
$search_pattern = ':^' . home_url( '/generate', 'relative' ) . ':';
```

`site_url()` points to where WordPress core files live (includes /wp/ on Bedrock), while `home_url()` is the public-facing URL — which is what REQUEST_URI actually contains.

See also #1897 which addresses the same class of Bedrock subdirectory issue on the JS side.

PR #2423, opened 2026-06-16
fix(faustwp): exclude updater code from wordpress.org build, add info.json

josephfusco:
Excludes the updater PHP files and info.json from the wordpress.org distribution via .distignore, clarifies the conditional require comment, qualifies the 1.5.0 self-update changelog entry, adds an explicit verifier assertion for updater file absence, and wires info.json refresh (version, last_updated, tested, download_link, versions map, changelog section) into scripts/versionPlugin.js so the non-wordpress.org channel manifest stays in sync at every version bump.

### DIFF
--- a/plugins/faustwp/.distignore
+++ b/plugins/faustwp/.distignore
@@ -23,3 +23,5 @@ package.json
 coverage
 phpstan.neon.dist
 /phpstan
+includes/updates
+info.json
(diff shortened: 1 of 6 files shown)


josephfusco:
Closing — initial audit was based on stale canary state and missed the existing composite action at .github/actions/release-plugin/action.yml, which already handles the two-build split and info.json refresh. The .distignore + info.json + versionPlugin.js changes here either duplicate that infra or actively break the main/internal zip. Reopening with a much smaller scope.

PR #2445, opened 2026-06-22
[SPCS-132] Update brace-expansion and js-yaml

theodesp:


colinmurphy (approved):

examples/next/custom-toolbar/package-lock.json
moonmeister, line 1: Custom toolbar is deprecated, last I checked...do we need to worry about updates?

theodesp: Security apparently thinks it's still a valid package:

https://github.com/advisories/GHSA-jxxr-4gwj-5jf2/dependabot?query=faustjs

maybe we can security to skip this check

